Issue #2, opened 2022-04-09
[IDEA] Message to be signed should be cleartext


bluesign:
Currently it is showing something like: 

`4:4163636570743a204170706c792061646a7573746d656e747320746f207472616e73616374696f6e2066656520706172616d65746572732074686572656279206368616e67696e67206665652063616c63756c6174696f6e73:1649521564405`

it should be more like: 

`Voting on Cast Proposal Number 4: Selected Option: Accept: Apply adjustments to transaction fee parameters thereby changing fee calculations`

Making user sign machine readable stuff is very bad practice.

PR #8, opened 2022-04-18
Support ledger tx (#2)

pgebheim:
* support connecting with ledger and signing tx for voting

* stubbed out backend validations, need support on getting blockId from grpc endpoint, need clear path about adding tx to_address to verify voting option

* add tx timestamp comparison

* pass in tx to addresses that map to vote choices

* add transaction id to message that goes to ipfs

* clean up


jacksonConrad (approved):

jacksonConrad:
Merged this into our private repo, and will be getting this repo up to date, so closing this out.

PR #13, opened 2022-05-18
feat(front): add inline images

germanurrus:
Screen Shot 2022-04-26 at 22 16 37

Screen Shot 2022-04-26 at 22 16 48

### DIFF
--- a/frontend/packages/client/src/components/UploadImageModal.js
+++ b/frontend/packages/client/src/components/UploadImageModal.js
@@ -48,7 +48,8 @@ function ImageUploader({
     uploadFile,
   ]);
 
-  const fileName = image?.name ?? image.file.name;
+  const fileName = image?.name ?? image?.file.name;
+
   const reducedFileName =
     fileName.length > 20 ? `${fileName.slice(0, 25)}...` : fileName;
   return (
--- a/frontend/packages/client/src/components/UploadImageModal.js
+++ b/frontend/packages/client/src/components/UploadImageModal.js
@@ -123,7 +123,6 @@ const UploadArea = ({ getRootProps, getInputProps, enableUpload }) => {
       )}
       {!enableUpload && (
         <div className="is-flex is-flex-direction-column is-align-items-center is-justify-content-center">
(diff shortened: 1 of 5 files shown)


frontend/packages/client/src/components/UploadImageModal.js
```diff
@@ -0,0 +1,325 @@
...
+    onUploaded,
+    loading,
+    error,
+    imageKey,
+    onUploadStared,
+    deleteImage,
+    uploadFile,
+  ]);
+
+  const fileName = image?.name ?? image.file.name;
```
CJamesKeller, line 51: Not a necessary change, but if we are checking whether `image` exists here in the first term, I would think the check should be applied in the second term, as well.

frontend/packages/client/src/components/UploadImageModal.js
```diff
@@ -0,0 +1,325 @@
...
+import { Upload, Bin } from "../components/Svg";
+import { Loader } from "../components";
+
+const IMAGE_STATUS = {
+  notStarted: "not-started",
+  uploading: "uploading",
+  uploaded: "uploaded",
+};
+
+function ImageUploader({
```
CJamesKeller, line 13: In general, I would favor splitting components into separate files and using folders to group them by functionality/larger components, with the top-level `components` folder being mainly for those which are shared across multiple areas. What do you think?

CJamesKeller (approved):
Looks good! A lot of functionality here -- Thank you for tackling this!

germanurrus:
@CJamesKeller I'm doing one last update on this PR: image uploads should be restricted to one and adding a caption below the image

frontend/packages/client/src/components/UploadImageModal.js
germanurrus: yes, I agree. We should take this approach...perhaps there might be some refactor to be done to achieve this on all the repo but I think it's convenient effort

germanurrus:
@CJamesKeller @dbslone updates done here. It's Ready for review: I'll create a second PR to adjusting the style for the caption text added at the bottom of the image.

dbslone (approved):
LGTM

PR #17, opened 2022-05-20
Eng-574 header; Eng-572 member count; Eng-568 join button; lint fixes

CJamesKeller:


frontend/packages/client/src/components/Community/JoinCommunityButton.js
```diff
@@ -0,0 +1,93 @@
...
+
+  const hoverToggle = () => {
+    if (joinedBtnText === "Joined") {
+      toggleJoinedBtnText("Leave");
+      setBorderColor("#FBD84D"); // defined in App.sass
+    } else {
+      toggleJoinedBtnText("Joined");
+      setBorderColor("transparent");
+    }
+  };
```
dbslone, line 37: Could we simplify the button so that we only have Join or Leave? This would simplify the logic needed and make the action of the button very clear without needing to hover to understand the action that will be performed if clicked.

frontend/packages/client/src/components/Community/JoinCommunityButton.js
```diff
@@ -0,0 +1,93 @@
...
+
+  const handleMembership = () =>
+    !isMember ? joinCommunity() : leaveCommunity();
+
+  const btnText = () => (!isMember ? "Join" : joinedBtnText);
+
+  return (
+    <>
+      {enableJoin && (
```
dbslone, line 70: What are your thoughts on making the responsibility of the component displaying the button to dermine if the button should render instead of the logic being in the button component?

```js
const Parent = () => {
  return (
     {enableJoin && <JoinCommunityButton />
  )
}
```

CJamesKeller: Yeah, that's probably a better approach, in case we have different cases come up in the future. I think I went too far in the directions of grouping it all. Will amend!

frontend/packages/client/src/components/Community/JoinCommunityButton.js
CJamesKeller: Confirmed we can do this, and implementing the change.

frontend/packages/client/src/components/Community/CommunityCard.js
```diff
@@ -101,9 +72,16 @@ const CommunityCard = ({
...
-    <Link to={`/community/${id}?tab=about`} style={{ color: "inherit" }}>
-      {Body}
-    </Link>
+    <>
+      <Link to={`/community/${id}?tab=about`} style={{ color: "inherit" }}>
+        {Body}
+      </Link>
+      <div
+        style={{ position: "absolute", margin: 0, top: "50px", right: "50px" }}
+      >
```
germanurrus, line 81: setting this one as absolute causes overlapping with long text communities:
maybe added as default and align it with bulma columns ?

Screen Shot 2022-05-23 at 12 04 54

frontend/packages/client/src/components/ProposalsList/ProposalList.sass
```diff
@@ -1,45 +1,89 @@
...
   border-top-right-radius: 8px
 .proposal-title
-  color: #000000 
-  font-weight: 400 
+  color: #000000
+  font-weight: 400
   font-size: 20px
   font-height: 24px
+.community-wrapper
+  padding-top: 0
```
germanurrus, line 18: maybe use pt-0 from bulma?

frontend/packages/client/src/components/ProposalsList/ProposalList.sass
```diff
@@ -1,45 +1,89 @@
...
+.community-header-wrapper
+  background-color: #F2F2F2
+  width: 100%
+  max-height: 248px
+.community-header
+  justify-content: space-between // here because bulma class didn't override
```
germanurrus, line 24: maybe using `is-justify-content-space-between`

frontend/packages/client/src/components/ProposalsList/ProposalList.sass
```diff
@@ -1,45 +1,89 @@
...
+  max-width: 85px
+.community-logo
+  border-radius: 50%
+.community-info
+  padding: 19.5px 24px
+  max-width: 225px
+.community-name
+  font-family: Arimo
+  font-size: 20px
+  font-weight: bold
```
germanurrus, line 39: if you don't define font-family it works since Aritmo is de primary: `$family-primary: "Arimo"`
font-size 20px can be replaced by `is-size-5` from bulma.
font-weight: bold can be replaced by `has-text-weight-bold` from bulma

frontend/packages/client/src/components/ProposalsList/ProposalList.sass
```diff
@@ -1,45 +1,89 @@
...
+.community-numbers
+  font-family: Arimo
+  color: #757575
+.community-content
+  padding-top: 31px
+.community-numbers
+  font-size: 14px
```
germanurrus, line 46: maybe group with https://github.com/DapperCollectives/CAST/pull/17/files#diff-b1cb312a16b0cd47b7accc8a5591f38dab55b81ed69d660504f54b1a7c76e5aaR40

germanurrus (commented):
looks good! left some comments to adjust the nav bar I think an update is necessary to show the join button on the community header here https://github.com/DapperCollectives/CAST/blob/bd83c1db491008771bb732d85d4b8558c6ed0e4b/frontend/packages/client/src/pages/Community.js#L136-L137

dbslone:
Screen Shot 2022-05-23 at 9 39 28 AM

Left padding for active proposal header needs to be adjusted

dbslone:
Screen Shot 2022-05-23 at 9 41 21 AM
Right padding on closed proposals needs to be adjusted

frontend/packages/client/src/components/Community/JoinCommunityButton.js
```diff
@@ -0,0 +1,68 @@
...
+}) {
+  const { createCommunityUser, deleteUserFromCommunity } = useJoinCommunity();
+  const { injectedProvider, user } = useWebContext();
+  const { addr } = user;
+  const memberState = useUserRoleOnCommunity({
+    addr,
+    communityId,
+    roles: ["member"],
+  });
+  const [isMember, setMemberState] = useState(memberState);
```
dbslone, line 17: `isMember` is being passed as a prop to the component but here you are using a state variable with the same name. 

After investigating further it seems that `isMember` is always set to false and the same is happening for `enableJoin`.  Are you able to update the API endpoints to return this data or do we need Richard to help?

CJamesKeller: Provided this information exists to be made use of, I can make those changes on Tuesday.

CJamesKeller: @dbslone , this is done

frontend/packages/client/src/utils.js
```diff
@@ -33,8 +33,10 @@ export const checkResponse = async (response) => {
   return response.json();
 };
 
-export const isNotEmptyArray = (array) =>
-  Array.isArray(array) && array.length > 0;
+export const isEmptyArray = (array) =>
+  Array.isArray(array) && array.length === 0;
+
+export const isNotEmptyArray = (array) => !isEmptyArray(array);
```
germanurrus, line 40: Is this isNotEmptyArray used anywhere?

frontend/packages/client/src/hooks/useUserRoleOnCommunity.js
```diff
@@ -1,27 +1,35 @@
...
 
 export default function useUserRoleOnCommunity({
   addr,
   communityId,
   roles = [],
 } = {}) {
-  const { data: communityUser, loading } = useUserCommunities({
-    addr,
+
+  const [hasRole, setHasRole] = useState();
```
germanurrus, line 11: this is an extra state that is not necessary, the hook `useUserCommunities` provides a reusable state that holds the data we need to pass as a result of the `useUserRoleOnCommunity` and the useUserCommunities already provides loading state and it triggers it self so the useEffect wouldn't be necessary

frontend/packages/client/src/hooks/useUserRoleOnCommunity.js
```diff
@@ -1,27 +1,35 @@
...
+      setHasRole(included);
+    }
+  }, [communityId, loading, roles, userCommunities]);
+
+  useEffect(() => {
+    getUserRole();
+  }, [getUserRole]);
 
-  return roles.every((role) => rolesInCommunity.includes(role));
+  return hasRole;
```
germanurrus, line 34: I think reusing the old code and moving the loading variable should work, then a state and a useEffect wound't be necessary:

```
 const { data: userCommunities, loading } = useUserCommunities({
    addr,
    count: 100,
  })
  if (loading) {
    return null;
  }
  if (
    addr === null ||
    (Array.isArray(userCommunities) && userCommunities.length === 0 && !loading)
  ) {
    return false;
  }

  const rolesInCommunity =
    userCommunities
      ?.filter((datum) => datum.id.toString() === communityId.toString())
      .map((community) => community?.membershipType) ?? [];
  return roles.every((role) => rolesInCommunity.includes(role));
```

frontend/packages/client/src/pages/Community.js
```diff
@@ -156,6 +158,17 @@ export default function CommunityPage() {
...
+  const {
+    pagination: { totalRecords },
+  } = useCommunityMembers({ communityId });
+  const [totalMembers, setTotalMembers] = useState();
+  useEffect(() => {
+    setTotalMembers(totalRecords);
+  }, [totalRecords]);
+
+  // for now allowList just returns admin addresses
+  const { data: adminAddrs } = useAllowlist();
```
germanurrus, line 170: you could reuse the return value from there: https://github.com/DapperCollectives/CAST/pull/17/files#diff-23488fb68fdc6db2dd3f17eab17a1a0858d02c54d55aa5852e7479b5bb385fc3R156
 `const { data: admins } = useCommunityUsers({ communityId, type: "admin" });`

germanurrus (changes requested):
Thanks Chris for the updates! I've added some comments. 
I didn't see the changes for refetching roles as we talked, but I think the current implementation is a better approach and saves the extra query to update the roles after the user has join or left the community.

frontend/packages/client/src/components/Community/JoinCommunityButton.js
```diff
@@ -0,0 +1,70 @@
...
+      style={{ minWidth: "117px" }}
+    >
+      <button
+        className="button is-uppercase is-fullwidth"
+        style={{
+          backgroundColor: "black",
+          borderRadius: "200px",
+          color: "white",
+        }}
+        onClick={isMember ? leaveCommunity : joinCommunity}
```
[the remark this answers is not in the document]
CJamesKeller: I realized this is why I had initially had the button set with `position: absolute`. I will go back to that, but ensure the column is still set for the text.

frontend/packages/client/src/components/ProposalsList/ProposalList.sass
```diff
@@ -1,45 +1,69 @@
...
+    padding-top: 40px
 
-.proposal-header-spacing
-  min-height: 66px
-  padding-right: 48px !important
-  padding-left: 46px !important
-.proposal-status
-  padding-right: 0 !important
-  code span
-    margin-right: 0 !important
```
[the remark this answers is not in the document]
CJamesKeller: Added back, but at the top section, for clarity.

frontend/packages/client/src/hooks/useJoinCommunity.js
```diff
@@ -0,0 +1,112 @@
...
+      const { signUserMessage } = currentUser();
+      dispatch({ type: "PROCESSING" });
+      const timestamp = Date.now().toString();
+      const hexTime = Buffer.from(timestamp).toString("hex");
+      const url = `${process.env.REACT_APP_BACK_END_SERVER_API}/communities/${communityId}/users`;
+      const _compositeSignatures = await signUserMessage(hexTime);
+      if (_compositeSignatures.indexOf("Declined:") > -1) {
+        dispatch({ type: "SUCCESS" });
+        return { success: false };
+      }
```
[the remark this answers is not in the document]
CJamesKeller: Thank you for catching this change!

frontend/packages/client/src/utils.js
CJamesKeller: Not anymore -- removed.

frontend/packages/client/src/hooks/useUserRoleOnCommunity.js
CJamesKeller: That is true... I'm not sure where I got caught in an intermediate step, but reverting these changes after confirming everything works without them. Thanks!

CJamesKeller:
@germanurrus , @dbslone , have added changes for comments

frontend/packages/client/src/pages/Community.js
```diff
@@ -233,8 +230,8 @@ export default function Community() {
                   </span>
                 </p>
                 <div className="is-flex">
-                  {adminAddrs
-                    ? adminAddrs.slice(0, 5).map((adminAddr, idx) => (
+                  {admins
+                    ? admins.slice(0, 5).map((adminAddr, idx) => (
```
germanurrus, line 234:
```suggestion
  ? admins.slice(0, 5).map(({ addr: adminAddr }, idx) => {
```
admins is an array of objects that needs to use addr, otherwise the icons do not render correctly
Screen Shot 2022-06-06 at 13 06 45

germanurrus (changes requested):
Thanks Chris! just left one question and one comment

frontend/packages/client/src/hooks/useJoinCommunity.js
```diff
@@ -60,22 +44,9 @@ export default function useJoinCommunity() {
...
-      const _compositeSignatures = await signUserMessage(hexTime);
-      if (_compositeSignatures.indexOf("Declined:") > -1) {
-        dispatch({ type: "SUCCESS" });
-        return { success: false };
-      }
-      const compositeSignatures = getCompositeSigs(_compositeSignatures);
-      if (!compositeSignatures) {
-        return { error: "No valid user signature found." };
-      }
-
```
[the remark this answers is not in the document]
CJamesKeller: This was removed per @dbslone - A user should be able to join/leave without having to sign a message. Could you confirm, David?

frontend/packages/client/src/components/Community/CommunityCard.js
```diff
@@ -29,9 +30,31 @@ const CommunityCard = ({
...
+        top: joinBtnPositions.smaller.top,
+        right: joinBtnPositions.smaller.top - windowWidth + 500
+      });
+      setBottomSpacer(true);
+    } else if (windowWidth >= 500 && joinBtnTopRight.top !== joinBtnPositions.bigger.top && bottomSpacer) {
+      setJoinBtnTopRight(joinBtnPositions.bigger);
+      setBottomSpacer(false);
+    }
+  }, [joinBtnTopRight.top, windowWidth, bottomSpacer, joinBtnPositions]);
+
```
germanurrus, line 54: is it possible to achieve all of this positioning with flex containers?  on mobile columns and column classes, in addition using custom components for mobile or desktop could help

germanurrus: I've created this ticket to track the improvement https://linear.app/dappercollectives/issue/CAS-103/improve-render-for-join-community-button-on-communitycard

germanurrus (approved):

PR #19, opened 2022-05-23
feat(front): group proposals into two list when filter has "all" selected

germanurrus:
- Group active and pending proposals into one list 
- Group cancelled and closed proposals into one list 

Screen Shot 2022-05-23 at 16 08 37

### DIFF
--- a/frontend/packages/client/src/components/CommunityProposals.js
+++ b/frontend/packages/client/src/components/CommunityProposals.js
@@ -5,13 +5,12 @@ import { useCommunityProposalsWithVotes, useMediaQuery } from "../hooks";
 import { FilterValues } from "../const";
 import DropDownFilter from "./ProposalsList/DropdownFilter";
 import WrapperResponsive from "./WrapperResponsive";
-import omit from "lodash/omit";
 
 export default function CommunityProposals({ community = { id: 1 } }) {
   const notMobile = useMediaQuery();
 
-  const proposalFilterValues = Object.values(
-    omit(FilterValues, ["inprogress", "terminated"])
+  const proposalFilterValues = Object.values(FilterValues).filter(
+    (value) => "In Progress" !== value && "Terminated" !== value
   );
 
   const [filterValue, setFilterValues] = useState(FilterValues["all"]);
--- a/frontend/packages/client/src/components/CommunityProposals.js
+++ b/frontend/packages/client/src/components/CommunityProposals.js
@@ -5,7 +5,7 @@ import { useCommunityProposalsWithVotes, useMediaQuery } from "../hooks";
 import { FilterValues } from "../const";
 import DropDownFilter from "./ProposalsList/DropdownFilter";
 import WrapperResponsive from "./WrapperResponsive";
-import { omit } from "lodash";
+import omit from "lodash/omit";
 
 export default function CommunityProposals({ community = { id: 1 } }) {
   const notMobile = useMediaQuery();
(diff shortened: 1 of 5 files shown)


frontend/packages/client/src/components/CommunityProposals.js
```diff
@@ -5,11 +5,14 @@ import { useCommunityProposalsWithVotes, useMediaQuery } from "../hooks";
 import { FilterValues } from "../const";
 import DropDownFilter from "./ProposalsList/DropdownFilter";
 import WrapperResponsive from "./WrapperResponsive";
+import { omit } from "lodash";
```
dbslone, line 8: To prevent the entire lodash library from being loaded you can directly import methods to improve tree-shaking during the build process.

`import omit from 'lodash/omit'`

frontend/packages/client/src/components/CommunityProposals.js
```diff
@@ -5,11 +5,14 @@ import { useCommunityProposalsWithVotes, useMediaQuery } from "../hooks";
...
 
 export default function CommunityProposals({ community = { id: 1 } }) {
   const notMobile = useMediaQuery();
 
-  const proposalFilterValues = Object.values(FilterValues);
+  const proposalFilterValues = Object.values(
+    omit(FilterValues, ["inprogress", "terminated"])
+  );
```
dbslone, line 15: any specific reason for using `omit` vs using native `filter` here?

frontend/packages/client/src/components/CommunityProposals.js
germanurrus: removed lodash and using filter

frontend/packages/client/src/components/CommunityProposals.js
germanurrus: using filter instead

frontend/packages/client/src/pages/Proposal.js
```diff
@@ -379,7 +379,7 @@ export default function ProposalPage() {
       <section className="section">
         <div className="container">
           <WrapperResponsive extraClasses="mb-6" extraClassesMobile="mb-3">
-            <Link to={`/community/${proposal.communityId}`}>
+            <Link to={`/community/${proposal.communityId}?tab=proposals`}>
```
germanurrus, line 382: this is for when coming back from the proposal page, to land again on the proposal tab instead of the about tab which is the first tab

frontend/packages/client/src/components/CommunityProposals.js
```diff
@@ -54,10 +61,12 @@ export default function CommunityProposals({ community = { id: 1 } }) {
...
-            <DropDownFilter
-              value={filterValue}
-              filterValues={proposalFilterValues}
-              setFilterValues={setFilterValues}
+            <DropDown
+              defaultValue={{ value: "all", label: "All" }}
+              values={proposalFilterValues}
+              onSelectValue={(value) => {
+                setFilterValues(value);
+              }}
```
germanurrus, line 69: I realized that we have the component Dropdown that should replace  `DropDownFilter` component. Once we merge this feature branch for community-v2 DropDownFilter could be safely deleted. I did the changes to use Dropdown now so we don't create more dependencies to the old component.

germanurrus:
@dbslone @CJamesKeller something that I did not realized is that it might be confusing how this filter works, so when "all" is selected the UI should show two groups of proposals: active & pending and closed & cancelled. The API for this requests is paginated and allows filtering with one filter value, so for this particular filter value (all) we are hitting the backend with two requests one that brings `status=inprogress` => backend returns active and pending proposals; and another one `status=terminated` =>  => backend returns closed and cancelled proposals; these are shown in two groups:

Screen Shot 2022-05-24 at 13 19 33

for any other filter value: `active`, `closed`, `cancelled`, `pending`, the frontend is doing one request with the corresponding `status=[value]`

Screen Shot 2022-05-24 at 13 20 19

dbslone (approved):

PR #21, opened 2022-05-24
Fix toggle of Proposal Choice

dbslone:
Bug: On the create proposal page when you click Visual then click Text-based the UI will display 2 empty text choices

### DIFF
--- a/frontend/packages/client/src/components/ProposalCreate/StepOne/TextBasedChoices.js
+++ b/frontend/packages/client/src/components/ProposalCreate/StepOne/TextBasedChoices.js
@@ -1,5 +1,6 @@
 import React, { useEffect } from "react";
 import { Plus, Bin } from "components/Svg";
+import { getProposalType } from "utils";
 
 const TextBasedChoices = ({
   choices = [],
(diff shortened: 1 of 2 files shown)


frontend/packages/client/src/components/ProposalCreate/StepOne/TextBasedChoices.js
```diff
@@ -10,9 +9,7 @@ const TextBasedChoices = ({
   initChoices,
 } = {}) => {
   useEffect(() => {
-    if (getProposalType(choices) !== "text-based") {
-      initChoices([]);
-    }
+    initChoices([]);
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
germanurrus, line 13: I did a test on this approach and when going back from the second step, it deletes the text options already saved in the first step. 
The issue I think that the `getProposalType` its not returning `image` when clicking on text-based after being in image-based. 

Changing it into this code below solves the problem:
````
export const getProposalType = (choices) => {
  if (
    choices.length === 2 &&
    choices.every((choice) => choice.choiceImgUrl !== undefined)
  ) {
    return "image";
  }
  return "text-based";
};

File: frontend/packages/client/src/components/ProposalCreate/StepOne/TextBasedChoices.js [hunk 2]
```diff
@@ -10,6 +10,7 @@ const TextBasedChoices = ({
   initChoices,
 } = {}) => {
   useEffect(() => {
+    console.log({ type: getProposalType(choices) });
```
germanurrus, line 13: this one can be removed

germanurrus (approved):
I left just one comment, thank you for the fix!

PR #23, opened 2022-05-25
Rich/cas 65

rvmelkonian:
added strategy 'one-token-one-address' with standard strategy tests. Weight is set to `1.00` per vote


backend/main/strategies/staked_token_weighted_default.go
```diff
@@ -0,0 +1,63 @@
+package strategies
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/brudfyi/flow-voting-tool/main/models"
```
dbslone, line 7: should this be importing from DapperCollectives/CAST instead?

backend/migrations/000001_init_db.up.sql
```diff
@@ -71,6 +71,8 @@ INSERT INTO voting_strategies (key, name, description)
 VALUES ('token-weighted-default', 'Token Weighted', 'Number of tokens held directly corresponds to the weight of the vote');
 INSERT INTO voting_strategies (key, name, description)
 VALUES ('token-weighted-capped', 'Token Weighted - Capped', 'Number of tokens held directly corresponds to the weight of the vote, up to the specified cap.');
+INSERT INTO voting_strategies (key, name, description)
+VALUES ('one-address-one-vote', 'One Address One Vote', 'one address is simply only allowed one vote, assets do not come into play.');
```
dbslone, line 75: Should this be in a new migration instead of altering an existing?

backend/main/strategies/staked_token_weighted_default.go
rvmelkonian: actually no in the go.mod file the package name is `module github.com/brudfyi/flow-voting-tool` so I can only import from this directory within the project. However, it's a good thing to raise @jacksonConrad, should we change the go.mod package directory?

backend/migrations/000001_init_db.up.sql
rvmelkonian: hmm I'm not sure, if I don't insert this strategy into the DB the tests wont pass.

rvmelkonian: I could add it into a separate migration file ... not sure what our pattern should be for this.

jacksonConrad: yeah it should be added as a separate migration, using the process outlined in the `README`

because when we `migrate up` on prod, only new migrations will be run.

backend/migrations/000014_add_one_address_strategy.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS voting_strategies;
```
dbslone, line 1: We shouldn't drop the entire table in the down. Instead this should just undo whatever is defined in the up.
Instead this should be:
```
DELETE FROM voting_strategies WHERE key='one-address-one-vote' LIMIT 1
```

rvmelkonian: got it. Thanks @dbslone, new to these migrations.

backend/migrations/000001_init_db.up.sql
```diff
@@ -59,7 +59,7 @@ CREATE TABLE votes (
 );
 
 -- Strategies
-CREATE TYPE strategies AS enum ('token-weighted-default', 'token-weighted-capped');
+CREATE TYPE strategies AS enum ('token-weighted-default', 'token-weighted-capped', 'one-address-one-vote');
```
jacksonConrad, line 62: we should remove this and add the enum type in the new up migration

backend/migrations/000014_add_one_address_strategy.up.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO voting_strategies (key, name, description)
+VALUES ('one-address-one-vote', 'One Address One Vote', 'one address is simply only allowed one vote, assets do not come into play.');
```
jacksonConrad, line 2: add `one-address-one-vote` to strategies enum here

backend/migrations/000014_add_one_address_strategy.down.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM voting_strategies WHERE key='one-address-one-vote'
```
jacksonConrad, line 1: FYI you wont need to worry about remove enum here because removing single type from enum in postgres is not supported (just keep this file as is!)

jacksonConrad (approved):

PR #24, opened 2022-05-25
fix(Cas 73): link proposal creation to community

germanurrus:


frontend/packages/client/src/components/CommunityProposals.js
```diff
@@ -6,7 +6,7 @@ import { FilterValues } from "../const";
 import DropDown from "./Dropdown";
 import WrapperResponsive from "./WrapperResponsive";
 
-export default function CommunityProposals({ community = { id: 1 } }) {
```
germanurrus, line 9: This was hardcoded

frontend/packages/client/src/components/CommunityProposals.js
```diff
@@ -86,7 +86,7 @@ export default function CommunityProposals({ community = { id: 1 } }) {
         {notMobile && (
           <div className="columns m-0">
             <div className="column p-0 is-10">
-              <Link to="/proposal/create">
+              <Link to={`/proposal/create?communityId=${communityId}`}>
```
germanurrus, line 89: when proposal creation starts we need a way to pass the communityId from where the flow was triggered

frontend/packages/client/src/pages/ProposalCreate.js
```diff
@@ -55,6 +60,14 @@ export default function ProposalCreatePage() {
...
     }
 
+    if (!communityId) {
+      notifyError({
+        status: "No community information provided",
+        statusText:
+          "Please restart the proposal creation from the community page",
+      });
+      return;
+    }
```
germanurrus, line 70: if at the end of the wizard the communityId query param is not in the Url, the proposal won't be created and an error message will be shown.

PR #26, opened 2022-05-26
Apply migrations when booting server

dbslone:
Automatically run migrations when booting in dev environment

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -4,6 +4,7 @@ build:
 	GOARCH=amd64 GOOS=linux go build -o ${BINARY_NAME} ./main
 
 build_macos:
+	make dev_run_migrate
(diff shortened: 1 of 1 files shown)


backend/Makefile
```diff
@@ -34,6 +36,11 @@ testmigrateup:
 testmigratedown:
 	migrate -path ./migrations -database "postgres://postgres:admin@localhost:5432/flow_snapshot_test?sslmode=disable" -verbose down
 
+dev_run_migrate:
+	@echo Running migrations...
+	make --ignore-errors migrateup &>/dev/null
+	make --ignore-errors testmigrateup &>/dev/null
```
dbslone, line 42: I'm using `--ignore-errors` here because otherwise the output from migrate will show errors when there is no migrations to apply and I'm sending the output to null to prevent the bootup output from being noisy. Open to other suggestions on how we could do this.

PR #28, opened 2022-05-26
cas-30 Update Proposal Card

CJamesKeller:
Depends on https://github.com/DapperCollectives/CAST/pull/13


germanurrus:
I think this one will be blocked by this PR https://github.com/DapperCollectives/CAST/pull/13 I'll try to finish the updates as soon as possible

frontend/packages/client/src/components/ProposalsList/ProposalCard.js
```diff
@@ -0,0 +1,87 @@
...
+
+const ProposalCard = ({ pr, style = {} }) => {
+  const isNotMobile = useMediaQuery();
+  const isTabletOnly = useMediaQuery(mediaMatchers.tabletOnly);
+  const isDesktopOnly = isNotMobile && !isTabletOnly;
+
+  const cardRef = useRef();
+  const [cardWidth, setCardWidth] = useState();
+  useEffect(() => {
+    // used for wrapping proposal body text
```
CJamesKeller, line 18: Very much open to a better solution for this, if possible.

frontend/packages/client/src/components/ProposalsList/ProposalCardBody.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import { stripHtml } from "string-strip-html";
+
+const ProposalCardBody = ({ name, body, cardWidth, inlineImage }) => {
+  const cardPaddingOnly = 48;
+  const aLittleExtra = 2;
+  const noImageDefault = cardPaddingOnly + aLittleExtra;
```
CJamesKeller, line 7: Can change the naming here if too informal.

frontend/packages/client/src/components/ProposalsList/ProposalCardFooter.js
```diff
@@ -0,0 +1,146 @@
...
+    if (justMounted && (FilterValues[computedStatus] === FilterValues.closed)) {
+      async function _getVotingResults() {
+        return getVotingResults(id);
+      }
+      _getVotingResults();
+    }
+  }, [computedStatus, getVotingResults, id, justMounted]);
+
+  // prevents memory leak in above useEffect for getVotingResults
+  useEffect(() => () => { setJustMounted(false) }, []);
```
CJamesKeller, line 28: Open to better approaches, but was unable to resolve the memory leak without this.

frontend/packages/client/src/components/ProposalsList/ProposalCardFooter.js
```diff
@@ -0,0 +1,146 @@
...
+  }, [votingResults]);
+
+  const iconStatusMap = {
+    [FilterValues.active]: <Active />,
+    [FilterValues.closed]:
+      textDecision !== "" ? <CheckCircle width="15" height="15" /> : null,
+  };
+
+  const textDescriptionMap = {
+    [FilterValues.active]: `${diffDuration} left`,
```
CJamesKeller, line 64: Technically differs from the design, but more consistent and readable, in my opinion.

frontend/packages/client/src/components/ProposalsList/ProposalCardFooter.js
```diff
@@ -0,0 +1,146 @@
...
+    [FilterValues.closed]:
+      textDecision !== ""
+        ? `${textDecision} ${winCount !== 0 ? `(${millify(winCount)})` : ""}`
+        : "",
+  };
+
+  const statusLabelMap = {
+    [FilterValues.active]: <StatusLabel status="Active" voted={voted} rounder
```
CJamesKeller, line 72: Note: We don't have any set line-length wrapping rules, but I wrapped attributes that would go over 100 chars, for readability, in all files. Can change this if preferred.

frontend/packages/client/src/components/ProposalsList/ProposalCardFooter.js
```diff
@@ -0,0 +1,146 @@
...
+      className="proposal-status-label has-text-weight-bold has-text-black"
+    />,
+    [FilterValues.pending]: <StatusLabel status="Pending" voted={voted} rounder
+      color="has-background-grey-light"
+    />,
+    [FilterValues.closed]: <StatusLabel status="Cancelled" voted={voted} rounder />,
+    [FilterValues.cancelled]: <StatusLabel status="Cancelled" voted={voted} rounder />,
+  };
+
+  // gross section to handle styling
```
CJamesKeller, line 83: Open to better approaches. It is gross.

frontend/packages/client/src/components/ProposalsList/ProposalCardHeader.test.js
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import { cleanup, render } from "@testing-library/react";
 import "@testing-library/jest-dom";
 import renderer from "react-test-renderer";
-import ProposalHeader from "./ProposalHeader";
+import ProposalCardHeader from "./ProposalCardHeader";
```
CJamesKeller, line 5: **Be advised**: Component has been updated, but test has _not_.

frontend/packages/client/src/components/ProposalsList/ProposalList.sass
```diff
@@ -4,12 +4,23 @@
...
+  font-size: 18px
+  line-height: 120%
+.proposal-body-text
+  min-height: 92px
+  height: 92px
+.proposal-text-truncated
+  line-height: 140%
+  display: -webkit-box
+  -webkit-box-orient: vertical
+  -webkit-line-clamp: 3
```
CJamesKeller, line 16: To wrap and clip text in the proposal body.

CJamesKeller:
- Implements new proposal card layout for all viewport sizes.
- Extracts proposal card and sub-components into separate files, leaving as much as possible the same as it was previously.
- Adds `date-fns` and `string-strip-html` deps to parse dates and html.
- Applies lint fix.
- Removes unnecessary components.

frontend/packages/client/src/components/ProposalsList/ProposalCardFooter.js
germanurrus: maybe triggering only by checking votingResults has something?

frontend/packages/client/src/components/ProposalsList/ProposalCardFooter.js
```diff
@@ -0,0 +1,146 @@
...
+  let extraClasses1 = "";
+  let extraClasses2 = "";
+  let extraStyles1 = {};
+  let extraClasses3 = "";
+  let extraClasses4 = "";
+
```
germanurrus, line 89: there's a WrapperComponent that handles styles from mobile and desktop https://github.com/DapperCollectives/CAST/blob/c37ce0380c3b9e71a72f7c753aed1b8f29b3d0ef/frontend/packages/client/src/components/WrapperResponsive.js 
maybe can be used to simplify the if clause

frontend/packages/client/src/components/ProposalsList/ProposalCardFooter.js
```diff
@@ -0,0 +1,146 @@
...
+  const statusLabelMap = {
+    [FilterValues.active]: <StatusLabel status="Active" voted={voted} rounder
+      color="has-background-orange"
+      className="proposal-status-label has-text-weight-bold has-text-black"
+    />,
+    [FilterValues.pending]: <StatusLabel status="Pending" voted={voted} rounder
+      color="has-background-grey-light"
+    />,
+    [FilterValues.closed]: <StatusLabel status="Cancelled" voted={voted} rounder />,
```
germanurrus, line 79:
```suggestion
    [FilterValues.closed]: <StatusLabel status="Closed" voted={voted} rounder />,
```

frontend/packages/client/src/components/ProposalsList/ProposalCardBody.js
```diff
@@ -0,0 +1,39 @@
...
+    // default so initial result is 0 until actual width is known
+    maxWidth = inlineImage ? inlineImageDefault : noImageDefault;
+  } else {
+    maxWidth = cardWidth;
+  }
+  if (inlineImage) {
+    maxWidth -= inlineImageDefault;
+  } else {
+    maxWidth -= noImageDefault;
+  }
```
dbslone, line 23: Is this needed? When I open the inspector and remove the maxWidth style from the body and adjust the window size I didn't notice any issues.

frontend/packages/client/src/components/ProposalsList/ProposalCardFooter.js
dbslone: There is no need to wrap the custom hook for fetching voting results with useEffect. If you look at how `useVotingResults` is used in [ProposalInformation.js](https://github.com/DapperCollectives/CAST/blob/main/frontend/packages/client/src/components/ProposalInformation.js#L166-L167) you can just the following here to achieve the desired result:

```
const { data: votingResults } = useVotingResults(id);
```

frontend/packages/client/src/components/ProposalsList/ProposalCardFooter.js
dbslone: Anytime you find yourself in a situation where you use generic variable names like these it's a good signal to think a little more about your naming convention. Naming things is one of the hardest parts of programming. 

In this situation you could use names like `Container`, `Icon`, `Status`, `Content`, etc. Try to use a name that describes a little about the element where the class or style is being applied.

CJamesKeller:
@germanurrus @dbslone Removed width check in `ProposalCardBody`, removed redundant `useEffect` hook, corrected `statusLabelMap`, and implemented `<WrapperResponsive />`.

dbslone (approved):

frontend/packages/client/src/components/ProposalsList/ProposalCardHeader.js
```diff
@@ -0,0 +1,25 @@
...
+import { useUserRoleOnCommunity } from "hooks";
+import { useParams } from "react-router-dom";
+
+const ProposalCardHeader = ({ creatorAddr }) => {
+  const { communityId } = useParams();
+  const isAdmin = useUserRoleOnCommunity({
+    creatorAddr,
+    communityId,
+    roles: ["admin"],
+  });
```
germanurrus, line 13: I think this might not be efficient, and we can work it on a separate PR and [ticket](https://linear.app/dappercollectives/issue/CAS-88/improve-use-of-useuserroleoncommunity-for-proposal-list), but the userRoleOnCommunity is using the endpoint `/users/${addr}/communities?count=${count}&start=${start}` which will bring all the communities that the address is part of, then we are making the same request for all headers. I think we can move this out of the proposal card header, and pass it down as a prop. The request would be done inside `useCommunityProposalWithVotes` [here](https://github.com/DapperCollectives/CAST/blob/fb64ffde79e37522cf555c3deb98a2831988126c/frontend/packages/client/src/hooks/useCommunityProposalsWithVotes.js#L7-L8) this endpoint returns up to 100 elements, then at the top level we can filter which ids we have and set the prop as admin or not.

germanurrus, line 13: I think this might not be efficient, and we can work it on a separate PR, but the userRoleOnCommunity is using the endpoint `/users/${addr}/communities?count=${count}&start=${start}` which will bring all the communities that the address is part of, then we are making the same request for all headers. I think we can move this our of the proposal card header, and pass it down as a prop. The request would be done inside `useCommunityProposalWithVotes` [here](https://github.com/DapperCollectives/CAST/blob/fb64ffde79e37522cf555c3deb98a2831988126c/frontend/packages/client/src/hooks/useCommunityProposalsWithVotes.js#L7-L8) this endpoint returns up to 100 elements, then at the top level we can filter which ids we have and set the prop as admin or not.

frontend/packages/client/src/pages/Community.js
```diff
@@ -194,7 +194,7 @@ export default function CommunityPage() {
         {loading && <Loader fullHeight />}
         {!loading && (
           <div className="columns m-0 p-0">
-            <div className="column p-0">
+            <div className="column p-0" style={{ maxWidth: "100%" }}>
```
germanurrus, line 197: question here: was this breaking somehow that needed to be added?

germanurrus (changes requested):
Looks good! just left some comments

germanurrus:
@CJamesKeller not sure why but looks like some comments are duplicated, maybe github issues?

frontend/packages/client/src/pages/Community.js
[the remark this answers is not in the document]
CJamesKeller: This was probably initially due to the width calculation for the text wrapping. Removed.

frontend/packages/client/src/components/ProposalsList/ProposalCardHeader.js
CJamesKeller: Made a change for this in `<CommunityProposalList />`, and realized I had been using the wrong value. Let me know what you think.

frontend/packages/client/src/components/ProposalsList/ProposalCard.js
```diff
@@ -0,0 +1,75 @@
...
+    <Link to={`/proposal/${pr.id}`}>
+      <div
+        className="border-light rounded-sm mb-5 proposal-card transition-all"
+        style={{ overflow: "hidden", ...style }}
+      >
+        {src && <ImageContentMobileAndTablet />}
+        <MainContent />
+      </div>
+    </Link>
+  );
```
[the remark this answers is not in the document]
CJamesKeller: Good call -- updated the file structure for the Proposal List & Card.

frontend/packages/client/src/components/CommunityProposals.js
```diff
@@ -53,7 +53,7 @@ export default function CommunityProposals({ communityId = 1 } = {}) {
         loadingActiveProposals;
 
   return (
-    <div className={`columns ${classesContainer}`}>
+    <div className={`columns mx-0 ${classesContainer}`}>
```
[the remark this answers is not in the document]
CJamesKeller: This was probably initially due to the width calculation for the text wrapping. Removed.

germanurrus (changes requested):
it's looking good Chris! I made some comments, thanks for the changes

frontend/packages/client/src/components/CommunityProposalList.js
```diff
@@ -50,47 +46,24 @@ const CommunityProposalList = ({
...
-                    by {pr.creatorAddr}
-                  </p>
-                  <p className="has-text-grey small-text">{pr.description}</p>
-                </div>
-              </div>
-            </Link>
+            <ProposalCard
+              pr={pr}
+              key={i}
+              isAdminProposal={admins.includes(pr.creatorAddr)}
```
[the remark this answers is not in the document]
dbslone: `isAdminProposal` is not being used here so the prop can be removed

frontend/packages/client/src/components/CommunityProposalList.js
```diff
@@ -50,47 +46,24 @@ const CommunityProposalList = ({
...
-                <div className="px-6 py-5 proposal-body-spacing">
-                  <h4 className="proposal-title is-4 mt-1 mb-2">{pr.name}</h4>
-                  <p className="has-text-grey mb-4 small-text">
-                    by {pr.creatorAddr}
-                  </p>
-                  <p className="has-text-grey small-text">{pr.description}</p>
-                </div>
-              </div>
-            </Link>
+              isAdminProposal={admins.includes(pr.creatorAddr)}
```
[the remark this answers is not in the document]
dbslone: same as above, remove the prop

frontend/packages/client/src/components/CommunityProposalList.js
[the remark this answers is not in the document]
germanurrus: I think `isAdminProposal' is used as a prop for `pr` here so needs to be passed done as a field for `pr` or as a prop for the component: 
used here https://github.com/DapperCollectives/CAST/pull/28/files/6613c6fe7894a2ecf5317d4f794fc422b774be6c#diff-a66b5e3ef6d31d017d9ce500657909be134fdf3b1bf05ac3547c03d487cbf005R7
passed here: 
https://github.com/DapperCollectives/CAST/pull/28/files/6613c6fe7894a2ecf5317d4f794fc422b774be6c#diff-49e211276445b84a1fd28ba2576e9eddd238fd568ebd8c0db04630400363ccb1R17
from this main component: 
https://github.com/DapperCollectives/CAST/pull/28/files/6613c6fe7894a2ecf5317d4f794fc422b774be6c#diff-30382faeb0dc980e175c3fc2fececb79b5b0df137d195f465f905d4c66daf50aR6
maybe pass this as a separate prop form the main content so it's not lost in the prop down drilling?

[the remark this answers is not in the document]
CJamesKeller: Shouldn't have missed that! Updated.

germanurrus (approved):

PR #15, opened 2022-05-18
feat(front): add create community 

germanurrus:
Last step updated with new fields:

Screen Shot 2022-05-25 at 11 02 28

To test, start frontend and go to `http://localhost:3000/#/community/create`

Required fields so far are:
- First Step:
Image
Name
About
category
Terms 

- Second Step:
None, can be left with not address, creator will be added as admin, author and member

- Last step:
Contract address: can be any string (this will be updated)
Contract Name: can be any string (this will be updated)
Storage Path: can be any string (this will be updated)
Proposal threshold: a number
checkbox not required

Hit create and a new community will be created.

PR #29, opened 2022-06-01
Rich/cas 69

rvmelkonian:
**Community Creation: Enforce Token Threshold for Proposal Creation**

This PR adds functionality that allows communities to set a threshold of any fungible token on the flow blockchain. If a threshold is set, a user wishing to create a new proposal must have this balance or higher to submit successfully. For this to work we need the contract name, address and public path inputted from the frontend (from the user, community admin etc) 

I've added the necessary code to allow us to use that data and parse it into a cadence script. In order to get all these values written into the script dynamically, I implemented the function `replaceContractPlaceholders` inside `/shared/flow.go`, as well as `EnforceTokenThreshold`. The public path can be passes in as a regular script parameter.

The function `CadenceValueToInterface` comes from Overflow, it allows us to easily cast any cadence value into type `interface`. 

I also added an additional test inside `proposal_test.go` - `Should be able to create a valid proposal given a token threshold` as well as the required utility functions.


backend/main/community_test.go
```diff
@@ -52,6 +53,7 @@ func TestCreateCommunity(t *testing.T) {
 	communityStruct := otu.GenerateCommunityStruct("account")
 	communityPayload := otu.GenerateCommunityPayload("account", communityStruct)
 	response := otu.CreateCommunityAPI(communityPayload)
+	fmt.Printf("%+v\n", response.Body)
```
dbslone, line 56: Is this just a debug statement while you were testing or is this something we need in the test?

dbslone:
Not sure why but `.DS_STORE` is showing up in the diff. Seems to be in the top-level `.gitignore` file though so it's strange this is happening.

backend/main/proposal_test.go
```diff
@@ -1,4 +1,4 @@
-package main
+pCommunity Creation: Enforce Token Threshold for Proposal Creationackage main
```
dbslone, line 1: looks like there was a typo here

backend/migrations/000013_add_community_categories_and_toc.up.sql
```diff
@@ -22,4 +22,5 @@ INSERT INTO community_types (key, name, description)
 VALUES ('service', 'Service', 'Service Community');
 
 ALTER TABLE communities ADD COLUMN category community_type;
-ALTER TABLE communities ADD COLUMN terms_and_conditions_url VARCHAR(256);
\ No newline at end of file
+ALTER TABLE communities ADD COLUMN terms_and_conditions_url VARCHAR(256);
+ALTER TABLE communities ADD COLUMN only_authors_to_submit BOOLEAN;
```
dbslone, line 26: We shouldn't alter migrations that have already been applied to the database because the change here will never be applied to the production database. Instead these alter table commands should be performed in migration #15.

backend/main/community_test.go
rvmelkonian: actually we may still need this. When running tests Golang does print out all the error logging in the server. I was encountering a weird IPFS time out error. So I've submitted this PR now Jackson will test locally to see if he gets the IPFS error, (it seems to be intermittent). Will need this log here for now.

backend/main/proposal_test.go
rvmelkonian: thanks, will fix

backend/main/proposal_test.go
```diff
@@ -61,14 +61,14 @@ func TestCreateProposal(t *testing.T) {
...
-		response := otu.CreateProposalAPI(payload)
 
+		response := otu.CreateProposalAPI(payload)
 		CheckResponseCode(t, http.StatusCreated, response.Code)
 
 		var p models.Proposal
 		json.Unmarshal(response.Body.Bytes(), &p)
 
-		assert.NotNil(t, p.Cid)
+		//assert.NotNil(t, p.Cid)
```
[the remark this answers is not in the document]
rvmelkonian: this is related to the IPFS. When IPFS works, we get returned this `Cid` value. It will be uncommented for final merge.

backend/migrations/000013_add_community_categories_and_toc.up.sql
rvmelkonian: thanks! Will fix.

rvmelkonian:
I noticed a mistake here that tests not picking up inside `func (fa *FlowAdapter) EnforceTokenThreshold(c *Contract) `, 
where I do 
```
flowAddress := flow.HexToAddress(*c.Addr)
```
This should be address of the voter, not address of the contract to check against. Will fix.

rvmelkonian:
this is fixed now. I also added a new migration file to create the contract new columns in community. Test for `Should be able to create a valid proposal given a token threshold` is passing. 

The test `A community author should be able to cancel a pending proposal they have created` is failing inside `TestupdateProposal` I'm not sure if this is related to the new code or not ... checking atm.

rvmelkonian:
Just confirmed only reason why :
```
A community author should be able to cancel a pending proposal they have created
A community author should be able to cancel an active proposal
```
Is because of the IPFS timeout error I was receiving before.

dbslone:
> Just confirmed only reason why these tests were failing :
> 
> ```
> A community author should be able to cancel a pending proposal they have created
> A community author should be able to cancel an active proposal
> ```
> 
> Is because of the IPFS timeout error I was receiving before.

Piñata can be very flaky at times. I've noticed multiple times I think something is broken but its because of an IPFS error

backend/main/shared/flow.go
```diff
@@ -181,6 +196,57 @@ func (fa *FlowAdapter) UserTransactionValidate(address string, message string, s
...
+		return false, errors.New("balance is below threshold")
+	}
+
+	return true, nil
+}
+
+// Fungible Token Address here is hardcoded to emulator address, this should
+// be set based on environment
+func replaceContractPlaceholders(code string, c *Contract) []byte {
+	code = placeholderFungibleTokenAddr.ReplaceAllString(code, "0xee82856bf20e2aa6")
```
jacksonConrad, line 244: should remove the hardcoded address and replace with environment variable.

jacksonConrad: we should probably use the `flow.json` to define and fetch the contract address mappings based on environment.

... i'll brainstorm the best way to do this

backend/scripts/get_balance.cdc
```diff
@@ -0,0 +1,13 @@
+import FungibleToken from 0xee82856bf20e2aa6
```
jacksonConrad, line 1: I think it would be clearer if all these values were placeholders to get replaced. token contract name, address & public capability path

backend/main/shared/flow.go
```diff
@@ -181,6 +196,57 @@ func (fa *FlowAdapter) UserTransactionValidate(address string, message string, s
...
+
+	value := CadenceValueToInterface(cadenceValue)
+	balance, err := strconv.ParseFloat(value.(string), 64)
+	if err != nil {
+		log.Error().Err(err).Msg("error converting cadence value to float")
+		return false, err
+	}
+
+	if balance < *c.Threshold {
+		return false, errors.New("balance is below threshold")
```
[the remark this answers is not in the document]
rvmelkonian: ofc !

PR #30, opened 2022-06-02
implemented FetchBalance on strategies, refactored controllers

rvmelkonian:
In this PR I added code so that the strategies each implement a `FetchBalance` function

```FetchBalance(db *shared.Database, b *models.Balance, sc *shared.SnapshotClient) (*models.Balance, error)```

Before we were doing this logic in the controller. Now I have updated the code to call the same models function on struct `Balance` `GetBalanceByAddressAndBlockHeight`, I also moved the generic error and `if balance.ID == ""` block into the `FetchBalance`. This is beneficial as it reduces the length of the controller function `createVoteForProposal` and abstracts some logic away. 

However, currently, each strategy is implementing the exact same logic for Fetch Balance, and I can't imagine a case where they would be different. If they're not going to be unique for each strategy, we could move the logic from lines 14 to 35 into a reusable function inside `/shared` 

Additionally in other controller functions such as `getVotesForProposal` we are calling the models method on struct `Vote` `GetVotesForProposal` which already joins the balances to the votes in one sql query, I thought it best not to alter this.

### DIFF
--- a/backend/main/strategies/one_address_one_vote.go
+++ b/backend/main/strategies/one_address_one_vote.go
@@ -4,9 +4,35 @@ import (
 	"fmt"
 
 	"github.com/brudfyi/flow-voting-tool/main/models"
+	s "github.com/brudfyi/flow-voting-tool/main/shared"
+	"github.com/jackc/pgx/v4"
+	"github.com/rs/zerolog/log"
 )
 
 type OneAddressOneVote struct{}
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: backend/main/strategies/one_address_one_vote.go [hunk 1]
```diff
@@ -4,10 +4,36 @@ import (
 	"fmt"
 
 	"github.com/brudfyi/flow-voting-tool/main/models"
+	s "github.com/brudfyi/flow-voting-tool/main/shared"
```
dbslone, line 7: typo?

jacksonConrad:
Looks good!  I think the repeated code is fine for now.

PR #31, opened 2022-06-02
fix(cas-85): Add image caption styles on editor and on proposal page

germanurrus:
- Fix style of image caption on editor mode and on proposal review last step. Also on proposal view mode.
- Added a fix for showing html header tags like H1, H2 and so on by adding the `content` class by bulma 
- Remove adding a break line because content provides styles for block separation 

- How to test?
Go to `http://localhost:3000/#/proposal/create`

On Edit Mode: 
Screen Shot 2022-06-06 at 16 15 40

On View mode: 
Screen Shot 2022-06-06 at 16 19 04
Screen Shot 2022-06-06 at 16 23 11


frontend/packages/client/src/components/ProposalCreate/StepOne/index.js
```diff
@@ -8,17 +8,37 @@ import React, {
...
   EditorState,
-  ContentBlock,
   AtomicBlockUtils,
   Modifier,
+  ContentState,
+  DefaultDraftBlockRenderMap,
+  SelectionState,
+  RichUtils,
 } from "draft-js";
+import { Map } from "immutable";
```
dbslone, line 18: Is there a reason for using `Map` from immutable instead of using the native Map? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

frontend/packages/client/src/components/ProposalCreate/StepOne/index.js
```diff
@@ -242,27 +262,64 @@ const StepOne = ({
...
     );
+    // get the block with custom type and with text
+    const [updatedBlock] = csWithUpdatedBlock.getBlocksAsArray();
 
-    const newEditorState = EditorState.set(newEditorStateWidthImage, {
-      currentContent: contentStateWithCaption,
-    });
+    // add block updated and concat empty block at the end
+    const newContentState = ContentState.createFromBlockArray(
+      blockMapArray.concat(updatedBlock).concat(lastBlockAdded),
```
dbslone, line 307: instead of chaining methods I think you could just use `.concat(updatedBlock, lastBlockAdded)`

frontend/packages/client/src/components/ProposalCreate/StepOne/index.js
```diff
@@ -333,10 +409,11 @@ const StepOne = ({
             editorState={localEditorState}
             toolbarClassName="toolbarClassName"
             wrapperClassName="border-light rounded-sm word-break-all"
-            editorClassName="px-4"
+            editorClassName="px-4 content"
```
germanurrus, line 412: this is a new class added because of this: https://bulma.io/documentation/elements/content/
elements like `h1` were being overwritten by not having the content class as a wrapper.

frontend/packages/client/src/components/ProposalCreate/StepThree.js
```diff
@@ -28,22 +27,16 @@ const StepThree = ({ stepsData, setStepValid }) => {
...
-    .replace(/(?:\r\n|\r|\n)/g, "<br>");
+  const htmlBody = customDraftToHTML(currentContent);
 
   return (
     <div>
       <ProposalStatus proposal={proposal} />
       <h1 className="title mt-5 is-3">{stepsData[0]?.title}</h1>
       <div
-        className="mt-6 mb-5 proposal-copy word-break-all"
+        className="mt-6 mb-5 proposal-copy word-break-all content"
```
germanurrus, line 39: same here, this is a new class added because of this: https://bulma.io/documentation/elements/content/

frontend/packages/client/src/pages/Proposal.js
```diff
@@ -454,7 +455,7 @@ export default function ProposalPage() {
                   >
                     {proposal.body && (
                       <div
-                        className="mt-4 mb-6 proposal-copy"
+                        className="mt-4 mb-6 proposal-copy content"
```
germanurrus, line 458: same here adding content class

frontend/packages/client/src/components/ProposalCreate/StepOne/index.js
germanurrus: updated

frontend/packages/client/src/App.sass
```diff
@@ -42,6 +42,7 @@ $border-color-light: #DBDBDB
 @import '~bulma/sass/elements/container.sass'
 @import '~bulma/sass/elements/progress.sass'
 @import '~bulma/sass/elements/title.sass'
+@import '~bulma/sass/elements/content.sass'
```
germanurrus, line 45: This was necessary to use the content class on the editor

frontend/packages/client/src/utils.js
```diff
@@ -1,3 +1,5 @@
+import { stateToHTML } from "draft-js-export-html";
```
germanurrus, line 1: replaces `draftjs-to-html` with `draft-js-export-html` since it provides better functions for block mapping and  creating html

frontend/packages/client/src/components/ProposalCreate/StepThree.js
```diff
@@ -28,22 +27,16 @@ const StepThree = ({ stepsData, setStepValid }) => {
...
-  const rawContentState = convertToRaw(
-    stepsData[0]?.description?.getCurrentContent()
-  );
-
-  const markup = draftToHtml(rawContentState);
+  const currentContent = stepsData[0]?.description?.getCurrentContent();
 
-  const htmlBody = markup
-    .replace(/target="_self"/g, 'target="_blank" rel="noopener noreferrer"')
-    .replace(/(?:\r\n|\r|\n)/g, "<br>");
```
germanurrus, line 39: remove and handled by https://github.com/DapperCollectives/CAST/pull/31/files#diff-36c9fb6b191cd8f048fcd99f91b906996a9db94bc2b63602e2c7452209af6190R122-R135

germanurrus:
@dbslone, @CJamesKeller changes done and ready for review

frontend/packages/client/src/components/ProposalCreate/StepOne/index.js
dbslone: Was looking into this further and looks like `draft-js` is using Immutable so we can't use the native map. After looking at our yarn.lock file we already have immutable package. If we update the package.json file to use: `immutable@^4.0.0` this will prevent us from having 3 different version of immutable included. Currently immutable doesn't support tree shaking so the entire package is being included so by pinning to v4.0 we can reduce the bundle size

frontend/packages/client/package.json
```diff
@@ -4,13 +4,14 @@
...
-    "@onflow/fcl": "^0.0.77",
     "@onflow/config": "0.0.2",
+    "@onflow/fcl": "^0.0.77",
     "@onflow/six-transfer-tokens": "0.0.8",
     "bulma": "0.9.3",
     "cookies-js": "^1.2.3",
     "draft-js": "^0.11.7",
-    "draftjs-to-html": "^0.9.1",
+    "draft-js-export-html": "^1.4.1",
+    "immutable": "^4.1.0",
```
dbslone, line 14:
```suggestion
    "immutable": "^4.0.0",
```

this will reduce the different versions of immutable included in our bundle see my note below for further details

dbslone (approved):

PR #33, opened 2022-06-06
created constructor for Proposal Results and new test case

rvmelkonian:
This is a bugfix for an issue that was not caught by out tests. We had no test that was hitting the endpoint ``http://localhost:5001/proposals/1/results``, this endpoint was returning an empty object, essentially null, in the case where a proposal with no votes was being fetched. 

A new constructor function has been added for ``ProposalResults`` struct and it is invoked inside ``TallyVotes``, the constructor sets default values for `Results_float` fields. 

A new test covering this scenario has been also been added.

### DIFF
--- a/backend/main/models/proposal_results.go
+++ b/backend/main/models/proposal_results.go
@@ -15,6 +15,20 @@ type ProposalResults struct {
 	Cid           *string            `json:"cid,omitempty"`
 }
 
+func NewProposalResults(id int) *ProposalResults {
+	p := new(ProposalResults)
+	p.Results = map[string]int{
+		"a": 0,
+		"b": 0,
+	}
+	p.Results_float = map[string]float64{
+		"a": 0.00,
+		"b": 0.00,
+	}
+	p.Proposal_id = id
+	return p
+}
+
 func (r *ProposalResults) GetLatestProposalResultsById(db *s.Database) error {
 	return pgxscan.Get(db.Context, db.Conn, r,
 		`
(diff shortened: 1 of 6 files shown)


backend/main/test_utils/vote_utils.go
```diff
@@ -26,6 +26,11 @@ func (otu *OverflowTestUtils) GetVotesForProposalAPI(proposalId int) *httptest.R
 	return otu.ExecuteRequest(req)
 }
 
+func (otu *OverflowTestUtils) GetResultsForProposalAPI(proposalId int) *httptest.ResponseRecorder {
+	req, _ := http.NewRequest("GET", "/proposals/"+strconv.Itoa(proposalId)+"/results", nil)
+	return otu.ExecuteRequest(req)
+}
```
dbslone, line 32: 👍 glad we have a test for this endpoint now

rvmelkonian: actually my bad, we did have a test hitting this endpoint, but not checking for the case of no votes! Just removing the duplicate code now

backend/main/models/proposal_results.go
```diff
@@ -15,6 +15,20 @@ type ProposalResults struct {
 	Cid           *string            `json:"cid,omitempty"`
 }
 
+func NewProposalResults(id int) *ProposalResults {
+	p := new(ProposalResults)
+	p.Results = map[string]int{
```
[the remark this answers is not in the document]
rvmelkonian: sure

PR #34, opened 2022-06-07
CAS-74: Update go.mod imports to github.com/DapperCollectives/CAST/backend

jacksonConrad:
update go.mod imports.  `github.com/brudfyi/flow-voting-tool` -> `github.com/DapperCollectives/CAST/backend`


dbslone (approved):
Looks good everything compiles and runs didn't notice any issues when testing the app

PR #35, opened 2022-06-07
fix: add author description 

germanurrus:
Waiting on comments form Product Team before merging

Add explanation when adding admin address they will be added as members and authors for the community:
-Added question mark with a popover: 

Screen Shot 2022-06-07 at 16 39 49

On desktop:
Screen Shot 2022-06-08 at 17 34 42

On Mobile could be:
Screen Shot 2022-06-08 at 17 35 08
Or:
Screen Shot 2022-06-08 at 17 45 26


frontend/packages/client/src/components/CommunityCreate/StepTwo.js
```diff
@@ -76,11 +103,28 @@ export default function StepTwo({
...
+            <button
+              className="delete has-text-grey rounded-full cursor-pointer"
+              style={buttonStyle}
+              onClick={openModalHelp}
+            >
+              ?
+            </button>
+          </>
+        }
+        description="The admins will be able to edit the space settings and moderate proposals. You must add one address per line. You must add one address per line."
```
dbslone, line 127: `You must add one address per line.` is duplicated at the end of the line

dbslone (changes requested):

frontend/packages/client/src/components/CommunityCreate/StepTwo.js
germanurrus: good catch! fixed

dbslone (approved):

CJamesKeller (approved):
Looks good! And it will be good to make use of the popover elsewhere, too

PR #37, opened 2022-06-08
fix(cas-95): member list shows duplicated addresses

germanurrus:
Description: 
On the community page, the member list shows duplicated addresses:
Screen Shot 2022-06-08 at 12 28 50

The endpoint used to retrieve data is `communities/{communityId}/users` the issue is that payload is arranged this way:
Screen Shot 2022-06-08 at 12 29 25
user has more than one role there will be as many entries by address as user types that address has. 
One possible solution would be to group data in the frontend, but since results are being paginated grouping needs to be done on every request and existing data in frontend state need to be used to group by address. 
In addition this endpoint is used to count community members, but since a user could have many roles, result may change depending on user roles per address. 

This PR changes the result payload for the endpoint `communities/{communityId}/users` to this:

Screen Shot 2022-06-09 at 10 59 25

also table has been updated to hide the last two columns:
Screen Shot 2022-06-10 at 13 56 55


backend/main/models/community_users.go
```diff
@@ -42,12 +64,12 @@ func GetUsersForCommunity(db *s.Database, communityId, start, count int) ([]Comm
...
+		return []CommunityUserType{}, 0, nil
 	}
 
 	// Get total number of users
 	var totalRecords int
-	countSql := `SELECT COUNT(*) FROM community_users WHERE community_id = $1`
+	countSql := `SELECT COUNT(*) FROM (SELECT addr FROM community_users WHERE community_id = $1 group BY community_users.addr) as temp_users_addr`
 	_ = db.Conn.QueryRow(db.Context, countSql, communityId).Scan(&totalRecords)
 
 	return users, totalRecords, nil
```
germanurrus, line 75: This function is used here: https://github.com/DapperCollectives/CAST/blob/958b9eea7f77c01aa3c3378da7df373d0d843da6/backend/main/server/app.go#L1498-L1499
the endpoint allows filtering, that with change will have two types of return payload. They question is if this should be implemented on a separate endpoint.

germanurrus:
@CJamesKeller this endpoint result is also affecting the members count on the community new header

germanurrus:
I pushed to style changes to hide `voting streak` and `score`

backend/main/models/community_users.go
jacksonConrad: i think it probably should be a separate endpoint, but we can merge this in as-is and i'll create a backlog ticket to refactor.

backend/main/models/community_users.go
```diff
@@ -29,11 +37,25 @@ type CommunityUserPayload struct {
...
 }
 
-func GetUsersForCommunity(db *s.Database, communityId, start, count int) ([]CommunityUser, int, error) {
-	var users = []CommunityUser{}
+func GetUsersForCommunity(db *s.Database, communityId, start, count int) ([]CommunityUserType, int, error) {
+	var users = []CommunityUserType{}
 	err := pgxscan.Select(db.Context, db.Conn, &users,
 		`
-		SELECT * FROM community_users WHERE community_id = $1
+		SELECT
```
jacksonConrad, line 44: there might be a more efficient way to write this query but i'm not totally sure off the top of my head.

we can revisit later if its a problem. :)

jacksonConrad (approved):

backend/main/models/community_users.go
germanurrus: yeah I agree, we could have functions running each of the CASE statements to simply the query and maybe creating a view or a materialized view could help too.

PR #39, opened 2022-06-08
cas 89 & 71 Balance of NFTs Strategy - Script implementation

rvmelkonian:
**Need to Resolve ...**
For some reason, when running executable `go run main.go`, there will be an error caused by line 62 inside /shared/flow.go
``content, err := ioutil.ReadFile("./flow.json")``
However, when you run tests, no error. 

If you change the line to : 
``content, err := ioutil.ReadFile("../flow.json")`` running the executable loads up the server, but then the tests throw an error. 

Need to consult @jacksonConrad about the best way to resolve this

**Description**

This PR adds implements a new strategy `balance-of-nfts`. If an nft matches the ``proposal_id`` vote weight is calculated. Pretty big updates had to made to the test suite as I needed to deploy ``NonFungibleToken``, ``MetadataViews`` and ``ExampleToken`` then actually mint some tokens with different accounts and seed the database, in order to test the script on-chain. 

We're using a new generalised script ``get_nfts_ids.cdc`` and replacing the dynamic values we need with regex, same as before. 

Now that all of this has been set up, it should be quite easy now to add variations of this strategy, in other words, calculate weight or voting power based on NFT data in different ways.

A new table has been added to the database ``nfts``
```
CREATE TABLE nfts (
  uuid UUID primary key,
  owner_addr VARCHAR(18) NOT NULL,
  id BIGINT NOT NULL,
  proposal_id BIGINT NOT NULL,
  created_at timestamp with time zone NOT NULL DEFAULT now(),
  updated_at timestamp with time zone NOT NULL DEFAULT now()
);
```

**Known Issues**

- If you want to run tests multiple times you'll run into an emulator error, telling you the contracts mentioned above have already been deployed to those addresses, so you have to cancel the running emulator, start it up again, then tests will run. 

- There are cases in cadence where an nft ID can either be an integer type or a string type ``uuid``. At the moment we are only handling ids of integer type, I can update this to handle both if required. 

**Improvements**
- Contract struct could be used inside Community struct. Can't remember why I didn't do this originally.


backend/main/server/app.go
```diff
@@ -568,9 +572,12 @@ func (a *App) createVoteForProposal(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	s.FlowAdapter = a.FlowAdapter
```
dbslone, line 575: I'm getting a compile error on this line: `main/server/app.go:575:4: s.FlowAdapter undefined (type Strategy has no field or method FlowAdapter)`

rvmelkonian: That should be fixed in the last push

rvmelkonian:
The test `should successfully create a vote for balance-of-nft strategy` inside `vote_test` is currently failing because of the error : `{"error":"user cannot vote on inactive proposal"}`

I just need to edit that test to make the proposal set to active; it should all check out after that!

backend/main/strategies/staked_token_weighted_default.go
```diff
@@ -36,16 +42,28 @@ func (s *StakedTokenWeightedDefault) FetchBalance(db *s.Database, b *models.Bala
...
+func (s *StakedTokenWeightedDefault) TallyVotes(
+	votes []*models.VoteWithBalance,
+	r *models.ProposalResults,
+) (models.ProposalResults, error) {
+
 	for _, vote := range votes {
-		p.Results[vote.Choice] += int(float64(*vote.StakingBalance) * math.Pow(10, -8))
-		p.Results_float[vote.Choice] += float64(*vote.StakingBalance) * math.Pow(10, -8)
+		if vote.StakingBalance == nil {
+			r.Results[vote.Choice] = 0.0
```
jacksonConrad, line 52: shouldnt be setting the results to 0 because that will erase any results that have been tallied. just need `continue`

backend/main/strategies/token_weighted_default.go
```diff
@@ -35,18 +42,29 @@ func (s *TokenWeightedDefault) FetchBalance(db *s.Database, b *models.Balance, s
...
-// func (s *TokenWeightedDefault) TallyVotes(votes []*models.VoteWithBalance, proposalId int) (models.ProposalResults, error) {
-func (s *TokenWeightedDefault) TallyVotes(votes []*models.VoteWithBalance, p *models.ProposalResults) (models.ProposalResults, error) {
+func (s *TokenWeightedDefault) TallyVotes(
+	votes []*models.VoteWithBalance,
+	p *models.ProposalResults,
+) (models.ProposalResults, error) {
 	//tally votes
 	for _, vote := range votes {
+		if vote.PrimaryAccountBalance == nil {
+			p.Results[vote.Choice] = 0.0
```
jacksonConrad, line 52: same here, just need `continue` in this if block

backend/main/test_utils/strategy_utils.go
```diff
@@ -84,3 +114,88 @@ func (otu *OverflowTestUtils) GenerateListOfVotes(proposalId int, count int) *[]
...
+		vote.Staking_balance = uint64(balance * 5)
+		vote.Block_height = uint64(0)
+
+		votes = append(votes, vote)
+	}
+
+	return &votes, nil
+}
+
+func (otu *OverflowTestUtils) CreateNFTVote(v models.Vote, ids []interface{}, contract *shared.Contract) VoteWithBalance {
```
jacksonConrad, line 155: any reason ids is type `[]interface{}` rather than `[]string` or `[]uint`?

backend/main/strategies/balance_of_nfts.go
```diff
@@ -0,0 +1,127 @@
...
+			return nil, err
+		}
+
+		for _, nftId := range nftIds {
+			nft := &models.NFT{
+				ID: nftId.(uint64),
+			}
+			vb.NFTs = append(vb.NFTs, nft)
+		}
+		err = models.CreateUserNFTRecord(db, vb)
```
jacksonConrad, line 55: before creating the NFT records, you should make sure that none of the NFT IDs have already been used to vote on this proposal. only create NFT records for the ones that havent been used.

backend/main/shared/flow.go
```diff
@@ -264,18 +287,60 @@ func (fa *FlowAdapter) EnforceTokenThreshold(creatorAddr string, c *Contract) (b
...
+	// In the case where ids are of string type we need an if statement somewhere to handle
+	nftIds := value.([]interface{})
+
+	return nftIds, nil
+}
+
+// @TODO
+// Hard coded Addresses here are for emulator dev only this should
+// be set based on environment var
+func replaceContractPlaceholders(code string, c *Contract, isFungible bool) []byte {
```
jacksonConrad, line 326: On the `main` branch this function looks like this:

```// Fungible Token Address here is hardcoded to emulator address, this should
// be set based on environment
func (fa *FlowAdapter) ReplaceContractPlaceholders(code string, c *Contract) []byte {
	fungibleTokenAddr := fa.Config.Contracts["FungibleToken"].Aliases[fa.Env]

	code = placeholderFungibleTokenAddr.ReplaceAllString(code, fungibleTokenAddr)
	code = placeholderTokenName.ReplaceAllString(code, *c.Name)
	code = placeholderTokenAddr.ReplaceAllString(code, *c.Addr)
	return []byte(code)
}```

Can you update this function to use the `flow.json` to resolve the address based on the environment?  will need to add the addresses for testnet & mainnet to `flow.json`. you should be able to find some of those in the `flow.json` in the VESSEL repo if necessary

**Review by jacksonConrad — CHANGES_REQUESTED** [review]

**Comment by jacksonConrad** [comment]
> If you change the line to : `content, err := ioutil.ReadFile("../flow.json")`
> running the executable loads up the server, but then the tests throw an error.

I think we should assume that the app will be run via `make macrun`/`make run`, etc. Unless you can figure out how to resolve the proper path when using `go run main.go` and set the `CWD` env variable accordingly in `main.go`.

In general you should be able to solve this by using `os.Getenv("CWD")` which is a variable i'm setting/using in `main.go`/`main_test.go` to enable the use of relative paths.  When running `make macrun`, the base path is`CAST/backend` so CWD should simply be set to `./` and so the path for `flow.json` will be `ioutil.ReadFile(os.Getenv("CWD") + "./flow.json")`.   In `main_test.go`, the base path is `CAST/backend/main`, so i'm setting `CWD` to `../` so that it will still be `CAST/backend`.

File: backend/main/test_utils/strategy_utils.go (hunk 4 above)
Comment (rvmelkonian) [reply backend/main/test_utils/strategy_utils.go:155]: its the type that returns from the function `CadenceValueToInterface` which is used when parsing the returned values from the cadence script
...
case cadence.Dictionary:
   result := map[string]interface{}{}
```

rvmelkonian:
ok this all updated. I had to update the strategy tests, both the tally functions and also within the tests itself as we are now using the constructor pattern in the endpoint with `NewProposalResults` this cleaned up the code inside the strategy `TallyVotes`` function. 

For the path, I used `APP_ENV` with an if statement before I saw your comment. It works fine, but lmk if you wana change it.

PR #43, opened 2022-06-10
CAS-85 - Proxy Analytics

germanurrus:
- updates on Plausible script

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -2,7 +2,7 @@
 
 ## Prerequisites
 
-- [GoLang 1.6](https://golang.org/doc/install)
+- [GoLang 1.16+](https://golang.org/doc/install)
 - [PostgreSQL 14.1](https://www.postgresql.org/download/)
 - [Flow CLI](https://docs.onflow.org/flow-cli/install/)
   - Note: See below for how install v0.30.2 (required)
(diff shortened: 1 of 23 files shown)


frontend/packages/client/public/index.html
```diff
@@ -28,8 +28,9 @@
     <title>CAST</title>
     <script
       defer
+      data-api="/api/event"
       data-domain="cast.fyi"
-      src="https://plausible.io/js/plausible.hash.js"
+      src="https://data.brud.dev/js/script.js"
```
dbslone, line 33: Wouldn't this only use our dev environment based on the url provided here? Is plausible being included as a package somewhere? Not sure I understand the change being made here.

germanurrus: My bad, I didn't add any explanation about this change. This is related to this: https://plausible.io/docs/proxy/introduction and could be possible that our plausible integration gets blocked by adblockers. 
At the end, the scripts running are the same it only changes the location from where we are pulling them from to run and make the adblocker "believe" that is one or our own scripts running.

dbslone: I think we should do this on the prod domain so it's `data.cast.fyi/script.js`. Can you also work with @fersan1985 to set up the DNS.

fersan1985: Hey @germanurrus or @dbslone That data.brud.dev where does it point to? Our brud.dev is a wildcard that we use in our ingress to point traffic to a specific service.

rvmelkonian (approved):

germanurrus:
Changes where deployed to staging directly https://github.com/DapperCollectives/CAST/pull/191

PR #47, opened 2022-06-15
fix(frontend): update role check for cancelling proposal

germanurrus:
This PR fixes showing the cancel button on proposals for ex-authors (after the user has left the community: all roles are removed when a user leaves the community). 
Before the check as done by looking if the user address was the same as the one that created the proposal. 
Since now we have roles in the community, only community authors should be able to chancel them. If the creator left the community, the user will need to rejoin and get an author role to be able to cancel a proposal

### DIFF
--- a/frontend/packages/client/src/hooks/useUserRoleOnCommunity.js
+++ b/frontend/packages/client/src/hooks/useUserRoleOnCommunity.js
@@ -36,7 +36,7 @@ export default function useUserRoleOnCommunity({
     return false;
   }
 
-  if (loading || roles.length === 0) {
+  if (loading || roles.length === 0 || !communityId) {
     return null;
   }
   if (pagination.next > 0) {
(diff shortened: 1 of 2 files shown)


frontend/packages/client/src/pages/Proposal.js
```diff
@@ -99,7 +104,12 @@ export default function ProposalPage() {
     error: strategiesError,
   } = useVotingStrategies();
 
-  const isAdmin = proposal && proposal?.creatorAddr === user?.addr;
```
germanurrus, line 102: creator of proposal was able to cancel it, now this has switch to the role the user has within the community.

frontend/packages/client/src/hooks/useUserRoleOnCommunity.js
```diff
@@ -36,7 +36,7 @@ export default function useUserRoleOnCommunity({
     return false;
   }
 
-  if (loading || roles.length === 0) {
+  if (loading || roles.length === 0 || !communityId) {
```
germanurrus, line 39: adding this check bc communityId could be undefined or null if called with params that are results of other hooks

dbslone (approved):

PR #51, opened 2022-06-16
feat: create addButton component

germanurrus:
- The add...something button is present in some components. Creating a stand alone component to avoid code duplication
Screen Shot 2022-06-16 at 17 29 53

to be used in https://linear.app/dappercollectives/issue/CAS-117/strategy-configuration-flow


frontend/packages/client/src/components/Community/CommunityEditorDetails.js
```diff
@@ -131,18 +131,12 @@ export const CommunityUsersForm = ({
...
-          loadingUsers ? "is-disabled has-text-grey" : "cursor-pointer"
-        }`}
-        onClick={onAddAddress}
-        style={loadingUsers ? { opacity: 0.5 } : {}}
-      >
-        <Plus />{" "}
-        <span className="ml-2 small-text is-flex is-align-items-center">
-          Add{` ${addrType}`}
-        </span>
-      </div>
```
germanurrus, line 145: Screen Shot 2022-06-16 at 17 34 24
also for authors but component is being reused so used only once

CJamesKeller (approved):

PR #57, opened 2022-06-18
cas-83 StepByStep forward backward navigation

CJamesKeller:
Includes update to `react-router-dom`. Only functional change is in `components/StepByStep/index.js`, where navigation is handled, with validation run once per page.


frontend/packages/client/src/components/CommunityProposals.js
```diff
@@ -70,7 +70,7 @@ export default function CommunityProposals({ communityId = 1, admins } = {}) {
           </WrapperResponsive>
           {!notMobile && (
             <div className="column p-0 mt-2 pl-2">
-              <Link to="/proposal/create">
+              <Link to={`../../proposal/create?communityId=${communityId}`}>
```
germanurrus, line 73: good catch!

frontend/packages/client/src/pages/Community.js
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from "react";
-import { useHistory, useParams, Link } from "react-router-dom";
+import { useSearchParams, useParams, Link } from "react-router-dom";
```
germanurrus, line 2: nice this comes out of the box!

frontend/packages/client/src/components/StepByStep/index.js
```diff
@@ -32,25 +38,80 @@ function StepByStep({
...
+  useEffect(() => {
+    const displayStep = searchParams.get("step");
+    if (once && (displayStep !== SPCopy)) { // navigated to a new step
+      setSPC(displayStep);
+      setOnce(false);
+    }
+  }, [SPCopy, once, searchParams]);
+
+  const goToStep = useCallback((_step, replace = false) => {
+    const step = _step + 1;
```
dbslone, line 72: When possible the pattern of `var` and `_var` should be avoided. In this usage an example of a better name would be `nextStep`

frontend/packages/client/src/components/StepByStep/index.js
```diff
@@ -32,25 +38,80 @@ function StepByStep({
...
+    let newQuery;
+    const displayStepLoc = search.indexOf("step=");
+    if (displayStepLoc === -1) { // no display step yet
+      newQuery = search.concat(`&step=${step}`);
+    } else {
+      newQuery = search.slice(0, displayStepLoc + "step=".length).concat(step);
+    }
```
dbslone, line 79: This could be simplified with the usage of `URLSearchParams` ie
```js
const q = new URLSearchParams(search)
q.set('step', x)
return q.toString()
```

dbslone (changes requested):

CJamesKeller:
@dbslone , @germanurrus , this is ready for review

frontend/packages/client/src/components/StepByStep/index.js
```diff
@@ -73,6 +59,71 @@ function StepByStep({
...
+  }, [currentStep, goToStep]);
+
+  useEffect(() => {
+    const displayStep = searchParams.get('step');
+    const displayStepNum = Number(displayStep);
+    const validStepsVals = Object.values(validSteps);
+    const numValidSteps = validStepsVals.length;
+    const allValidSteps = () =>
+      numValidSteps &&
+      validStepsVals.slice(0, displayStepNum - 1).every((val) => val === true);
```
germanurrus, line 100: not sure about this condition, it does not fail when displayStepNum is not a number because it's not evaluated due to the condition `numValidSteps` being 0 at first load. Same as in line 106. Could it be done more pure?

frontend/packages/client/src/components/StepByStep/index.js
```diff
@@ -73,6 +59,71 @@ function StepByStep({
...
+    goToStep,
+    searchParams,
+    steps.length,
+    isStepValid,
+    stepsData,
+    validSteps,
+    setValidSteps,
+    setCurrentStep,
+  ]);
+
```
germanurrus, line 126: Should this be split into two useEffects? 

- One for checking url is valid (it's a number and it falls under the bounds steps number) and
- A second one to check that all steps are valid (each step has valid data) and allow to move to the next/prior step the url is indicating

I think it's better to use multiple effects to separate concerns, useEffect is designed to handle only one concern.
When trying to use only one effect for multiple purposes, it decreases the readability 😢

germanurrus (changes requested):
thanks for the update! looks good, I left some comments

frontend/packages/client/src/components/CommunityProposalList.js
```diff
@@ -17,7 +17,7 @@ const EmptyPlaceHolder = ({ communityId }) => {
           <p className="has-text-grey mb-4 small-text">
             What will your community’s first proposal be?
           </p>
-          <Link to={`/proposal/create?communityId=${communityId}`}>
+          <Link to={`../../proposal/create?communityId=${communityId}`}>
```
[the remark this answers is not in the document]
CJamesKeller: I had not tried the more clean paths (I was thinking strictly like in the terminal) but this does seem to work in all cases, and has been adjusted!

frontend/packages/client/src/components/StepByStep/index.js
CJamesKeller: Updated to split out these two concerns.

germanurrus (changes requested):
Getting this loop error: 
```
index.js:1 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    at StepOne (http://localhost:3000/static/js/0.chunk.js:239:5)
    at div
    at div
    at section
    at StepByStep (http://localhost:3000/static/js/main.chunk.js:13121:95)
    at CommunityCreate (http://localhost:3000/static/js/0.chunk.js:1697:93)
    at Routes (http://localhost:3000/static/js/vendors~main.chunk.js:128837:5)
    at div
    at div
    at Suspense
    at AppPages
    at ErrorHandlerProvider (http://localhost:3000/static/js/main.chunk.js:18113:5)
    at Router (http://localhost:3000/static/js/vendors~main.chunk.js:128770:15)
    at HashRouter (http://localhost:3000/static/js/vendors~main.chunk.js:127611:5)
    at NotificationModalProvider (http://localhost:3000/static/js/main.chunk.js:18336:5)
    at Web3Provider (http://localhost:3000/static/js/main.chunk.js:18567:5)
    at ErrorBoundary (http://localhost:3000/static/js/vendors~main.chunk.js:123667:37)
    at App
````
 Looks like a cut condition is needed here: https://github.com/DapperCollectives/CAST/pull/57/files#diff-1f0ecdb68db063b3091e2549cbebf91f9015ff733286ab9737290fe49c7e0436R109

**Comment by CJamesKeller** [comment]
@germanurrus , how did you get that error state? I have tried all of the following, without error:

* List page --> Step 1 --> **back** to List page --> **forward** to Step 1
* List page --> Step 1 --> **manually** try to get to Step 3 (routed to Step 1)
* List page --> Step 1 --> **manually** try to get to non-existent Step "foo" (routed to Step 1)
* List page --> Step 1 --> Step 2 --> Step 3 --> www.google.com --> try to go **back** to Step 3 (routed to Step 1)

**Comment by germanurrus** [comment]
> @germanurrus , how did you get that error state? I have tried all of the following, without error:
> 
> * List page --> Step 1 --> **back** to List page --> **forward** to Step 1
> * List page --> Step 1 --> **manually** try to get to Step 3 (routed to Step 1)
> * List page --> Step 1 --> **manually** try to get to non-existent Step "foo" (routed to Step 1)
> * List page --> Step 1 --> Step 2 --> Step 3 --> [www.google.com](http://www.google.com) --> try to go **back** to Step 3 (routed to Step 1)

@CJamesKeller I was checking the create proposal flow and it works fine, the issue is with the create community flow: `http://localhost:3000/#/community/create?step=1` after hitting the "GET STARTED" button.  The community creation flow has a pre-step that is not considered a step on the step array.

File: frontend/packages/client/src/components/StepByStep/index.js [hunk 8]
```diff
@@ -73,6 +64,96 @@ function StepByStep({
...
+    const displayStep = searchParams.get('step');
+    const displayStepNum = !isNaN(Number(displayStep))
+      ? Number(displayStep)
+      : -1;
+    const validStepsVals = Object.values(validSteps);
+    const numValidSteps = validStepsVals.length;
+    if (
+      (!preStep &&
+        // prevent reaching a subsequent step until prior steps have data
+        ((displayStepNum > 1 && (!numValidSteps || !allValidSteps())) ||
```
germanurrus, line 132: `allValidSteps` needs to be called like this `allValidSteps(validStepsVals, displayStepNum)` otherwise it throws an error.

frontend/packages/client/src/components/StepByStep/index.js
```diff
@@ -73,6 +64,96 @@ function StepByStep({
...
+          // or in the case of a mismatch
+          (!Object.keys(stepsData).length && numValidSteps))) ||
+      (preStep &&
+        ((displayStepNum > 2 && (!numValidSteps || !allValidSteps())) ||
+          (!Object.keys(stepsData).length && numValidSteps > 1)))
```
germanurrus, line 137: This is causing an issue if there's a step that allows the user to not enter any data if it's not requited. 
I think we can remove checking if data is available as long as the step is valid. Data existence or validation is handled by each step and if it's ok then `isValidStep` is called. It can be tested on creating a community were step two adding admin and authors is empty it will cause to move to step 1 instead of step 3.

germanurrus (changes requested):
I left some comments, last changes broke the move to next step action

germanurrus:
Tested branch locally and works solid now @CJamesKeller.

rvmelkonian:
is this going to get merged in, or stale PR now?

dbslone:
it's stale now will close PR

PR #59, opened 2022-06-21
Update react version to 18 - Approach B

germanurrus:
Approach A: 
This PR will update: 
react to 18.2.0
react-dom to 18.2.0
react-scripts to 5.0.1
also remove `sass` from list of dependencies and added to dev dependencies. 

A couple of issues along the way: 
 - Webpack 5 no longer polyfills Node.js core modules automatically:

https://webpack.js.org/configuration/resolve/#resolvefallback

There’s an open PR that explains the situation: https://github.com/facebook/create-react-app/issues/11756

This approach use `patch-package` to patch `/node_modules/react-scripts/config/webpack.config.js` after package installing. https://github.com/facebook/create-react-app/issues/11756#issuecomment-1002637667 
This way we are adding necessary packages to the webpack bundle process used by `react-scripts`


frontend/packages/client/package.json
```diff
@@ -7,44 +7,45 @@
...
   },
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
     "test:update": "react-scripts test --updateSnapshot",
     "eject": "react-scripts eject",
-    "prettier": "npx prettier --write ."
+    "prettier": "npx prettier --write .",
+    "postinstall": "cd ../.. && yarn patch-package"
```
germanurrus, line 48: this script will adjust webpack config after any install done on the app

frontend/patches/react-scripts+5.0.1.patch
```diff
@@ -0,0 +1,17 @@
...
+       },
++      fallback: {
++        "url": require.resolve("url/"),
++        "http": require.resolve("stream-http"),
++        "https": require.resolve("https-browserify"),
++        "buffer": require.resolve("buffer/")
++      },
+       plugins: [
+         // Prevents users from importing files from outside of src/ (or node_modules/).
+         // This often causes confusion because we only process files within src/ with babel.
```
germanurrus, line 17: this is the patch file used to update webpack config

PR #65, opened 2022-06-22
CAS-135: Add Leaderboard Endpoint

jbluks:
**Ticket:** [CAS-135](https://linear.app/dappercollectives/issue/CAS-135/leaderbord-backend-v0)

**Description**

- Adds endpoint `/communities/:communityId/leaderboard` 
- Returns list of user addresses and score based on tallying votes on all proposals

### DIFF
--- a/backend/main/community_user_test.go
+++ b/backend/main/community_user_test.go
@@ -127,6 +127,34 @@ func TestGetCommunityUsersByInvalidType(t *testing.T) {
 	checkResponseCode(t, http.StatusBadRequest, response.Code)
 }
 
+func TestGetCommunityLeaderboard(t *testing.T) {
+	clearTable("communities")
+	clearTable("community_users")
+	clearTable("proposals")
+	clearTable("votes")
+	communityId := otu.AddCommunities(1)[0]
+	proposalIdA := otu.AddActiveProposals(communityId, 1)[0]
+	proposalIdB := otu.AddActiveProposals(communityId, 2)[0]
+	proposalIdC := otu.AddActiveProposals(communityId, 3)[0]
+	voteChoice := "a"
+
+	otu.CreateVoteAPI(proposalIdA, otu.GenerateValidVotePayload("user1", proposalIdA, voteChoice))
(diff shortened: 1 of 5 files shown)


rvmelkonian (changes requested):
- Can we create a new file `leaderboard_test.go` and put the tests in there? (I can see your reasoning here as the endpoint is under the `/communities/` api route, but we may need to add more leaderboard related tests if complexity grows.)

backend/main/models/community_users.go
```diff
@@ -99,6 +104,34 @@ func GetUsersForCommunityByType(db *s.Database, communityId, start, count int, u
...
+		SELECT v.addr, count(*) AS score FROM votes v 
+		JOIN proposals p ON p.id = v.proposal_id
+		WHERE p.community_id = $1
+		GROUP BY v.addr
+		ORDER BY score DESC
+		LIMIT $2 OFFSET $3
+		`, communityId, count, start)
+
+	// If we get pgx.ErrNoRows, just return an empty array
+	// and obfuscate error
```
rvmelkonian, line 120: can we remove these comments? I think they're getting duplicated around. The code here is self-explanatory.

backend/main/community_user_test.go
```diff
@@ -127,6 +127,34 @@ func TestGetCommunityUsersByInvalidType(t *testing.T) {
...
+	clearTable("community_users")
+	clearTable("proposals")
+	clearTable("votes")
+	communityId := otu.AddCommunities(1)[0]
+	proposalIdA := otu.AddActiveProposals(communityId, 1)[0]
+	proposalIdB := otu.AddActiveProposals(communityId, 2)[0]
+	proposalIdC := otu.AddActiveProposals(communityId, 3)[0]
+	voteChoice := "a"
+
+	otu.CreateVoteAPI(proposalIdA, otu.GenerateValidVotePayload("user1", proposalIdA, voteChoice))
```
rvmelkonian, line 141: from this line down to line 145.. Perhaps you could move these lines into their own `GenerateValidPayloads` function? And then we would need to call it only once inside this test. You could have params like `count` (how many times its being called) and `users` which could be an array of your user strings? 

We could then reuse it, being able to Generate multiple payloads will be useful for testing!

PR #71, opened 2022-06-23
fix: update validation on links form

germanurrus:
- On initial load save button as enabled although fields were not updated. Button should be enabled only if a field has changed from its original value

### DIFF
--- a/frontend/packages/client/src/components/Community/CommunityEditorDetails.js
+++ b/frontend/packages/client/src/components/Community/CommunityEditorDetails.js
@@ -289,7 +289,7 @@ const CommunityMembersEditor = ({
       submitComponent={
         <button
           style={{ height: 48, width: '100%' }}
-          className={`button transition-all is-flex has-background-yellow rounded-sm mt-5 is-uppercase is-${
+          className={`button transition-all is-flex has-background-yellow rounded-sm mt-5 is-uppercase is-uppercase is-${
             enableSave ? 'enabled' : 'disabled'
           }`}
           onClick={!enableSave ? () => {} : saveData}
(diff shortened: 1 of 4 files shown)


frontend/packages/client/src/components/Community/CommunityEditorDetails.js
```diff
@@ -289,7 +289,7 @@ const CommunityMembersEditor = ({
       submitComponent={
         <button
           style={{ height: 48, width: '100%' }}
-          className={`button transition-all is-flex has-background-yellow rounded-sm mt-5 is-uppercase is-${
+          className={`button transition-all is-flex has-background-yellow rounded-sm mt-5 is-uppercase is-uppercase is-${
```
germanurrus, line 292: change to uppercase the text

frontend/packages/client/src/components/Community/CommunityEditorLinks.js
```diff
@@ -129,11 +128,12 @@ export default function CommunityEditorLinks(props = {}) {
...
 
-  const { isValid } = useLinkValidator({ links, initialValues: props });
+  const { isValid, hasChangedFromOriginal } = useLinkValidator({
+    links,
+    initialValues: fields,
+  });
 
-  useEffect(() => {
-    setEnableSave(isValid);
-  }, [isValid]);
```
germanurrus, line 136: this wasn't necessary since a hook is re-rendering the object when fields are valid or are different from originals

sarahYounes-DapperLabs:
fix has been applied to every field except for the description field; but is still reproducible

sarahy-Dapper (commented):
looks great

sarahy-Dapper (approved):

PR #72, opened 2022-06-27
fix(frontend): fix user removal from community

germanurrus:
- When an admin or author user leaves a community by clicking the leave button. User loses all roles, if user happen to be admin or author these lists are not being updated and the user is still shown in the list as admin, author or both. 

Screen Shot 2022-06-26 at 22 11 58

- Also icon for admins do not get updated and still shows the admin icon: 
Screen Shot 2022-06-26 at 22 15 24

This PR adds a callback function that when user leaves the community and the user has an admin and/or author role, it triggers an update on hooks for listing admins and/or authors

### DIFF
--- a/frontend/packages/client/src/hooks/useCommunityUsers.js
+++ b/frontend/packages/client/src/hooks/useCommunityUsers.js
@@ -126,6 +126,12 @@ export default function useCommunityUsers({
     dispatch({ type: 'RESET_RESULTS' });
   };
 
+  // clears all results and pulls again
+  const reFetch = async () => {
+    resetResults();
+    await getCommunityUsers();
+  };
+
   useEffect(() => {
     getCommunityUsers();
   }, [getCommunityUsers]);
@@ -137,5 +143,6 @@ export default function useCommunityUsers({
     getCommunityUsers,
     fetchMore,
     resetResults,
+    reFetch,
   };
 }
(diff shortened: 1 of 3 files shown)


frontend/packages/client/src/hooks/useCommunityUsers.js
```diff
@@ -61,9 +61,9 @@ export default function useCommunityUsers({
     dispatch({ type: 'PROCESSING' });
     const url = `${
       process.env.REACT_APP_BACK_END_SERVER_API
-    }/communities/${communityId}/users?count=${count}&start=${start}${
-      type ? `&userType=${type}` : ''
-    }`;
+    }/communities/${communityId}/users${
+      type ? `/type/${type}` : ''
+    }?count=${count}&start=${start}`;
```
germanurrus, line 66: same here

frontend/packages/client/src/hooks/useCommunityUsers.js
```diff
@@ -40,9 +40,9 @@ export default function useCommunityUsers({
     dispatch({ type: 'PROCESSING' });
     const url = `${
       process.env.REACT_APP_BACK_END_SERVER_API
-    }/communities/${communityId}/users?count=${state.pagination.count}&start=${
+    }/communities/${communityId}/users${type ? `/type/${type}` : ''}?count=${
       state.pagination.start
-    }${type ? `&userType=${type}` : ''}`;
+    }&start=${state.pagination.start}`;
```
germanurrus, line 45: there was an endpoint update on the backend code that changed the url

sarahy-Dapper (approved):
looks good

PR #74, opened 2022-06-27
fix: make proposal description apear under stepper

germanurrus:
Editor container is zindex=1, so stepper needs to be higher: 

Screen Shot 2022-06-27 at 14 31 03

Screen Shot 2022-06-27 at 14 21 48


dbslone (approved):

sarahy-Dapper (approved):
looks good

PR #82, opened 2022-06-28
fix(frontend): fix autofocus on author editor step

germanurrus:
On community creation workflow, second step, autofocus is applied to last element on authors list. It should be applied to admins since it's the first input to complete that appears: 
Screen Shot 2022-06-28 at 11 09 15

Fixed: 
Screen Shot 2022-06-28 at 11 00 38

Next new inputs will have autofocus as the are added on the page

### DIFF
--- a/frontend/packages/client/src/components/Community/CommunityEditorDetails.js
+++ b/frontend/packages/client/src/components/Community/CommunityEditorDetails.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import { Bin, ValidCheckMark, InvalidCheckMark } from 'components/Svg';
 import { WrapperResponsive, Loader, AddButton } from 'components';
 import { useCommunityUsers } from 'hooks';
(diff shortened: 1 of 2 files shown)


frontend/packages/client/src/components/Community/CommunityEditorDetails.js
```diff
@@ -10,6 +10,15 @@ import useFlowAddrValidator, {
...
 
+const determineAutoFocus = (listLength, index, addr, autoFocusOnLoad) => {
+  return (
+    // fist element of list depends on configuration (autoFocusOnLoad)
+    (listLength === 1 && addr === '' && autoFocusOnLoad) ||
+    // new elements on list will be autofocus by default
+    (listLength === index + 1 && addr === '' && index !== 0)
+  );
+};
+
```
dbslone, line 21: Any way we could simplify this by using ref? Or does the current component structure prevent us from being able to take this approach?
```
useEffect(() => {
        mainRef.current.focus();
    }, [mainRef]);

<
 ```

File: frontend/packages/client/src/components/Community/CommunityEditorDetails.js [hunk 2]
```diff
@@ -80,6 +89,11 @@ export const CommunityUsersForm = ({
                     width: '100%',
                     ...(isInvalid ? {} : undefined),
                   }}
+                  ref={
+                    addrList.length === 1 && addr === '' && autoFocusOnLoad
+                      ? refOnFirstInput
+                      : undefined
+                  }
```
germanurrus, line 96: this component is used twice in the same step so the last focus will win. For that reason using autoFocusOnLoad to enable one and disable the second one. That way the first input will keep the focus.

sarahy-Dapper (approved):
issue persists on Firefox browser:
image

dbslone (approved):

PR #62, opened 2022-06-21
feat(frontend): update create community flow

germanurrus:
To test this PR go to `http://localhost:3000/#/community/create`

Add extra step on community creation to configure strategies available for the new community: 

Screen Shot 2022-06-27 at 10 57 49

PR #83, opened 2022-06-28
CAS-108 - validate contract address during community creation

germanurrus:
Depends on #62 
Added validation for flow addresses, when creating a community it will show an error modal if any of the addresses used is not a valid flow addr. Maybe wording and some style will need to be added: 

Screen Shot 2022-07-04 at 19 38 05

### DIFF
--- a/frontend/packages/client/src/pages/CommunityCreate.js
+++ b/frontend/packages/client/src/pages/CommunityCreate.js
@@ -17,6 +17,7 @@ export default function CommunityCreate() {
   const {
     user: { addr: creatorAddr },
     injectedProvider,
+    isValidFlowAddress,
   } = useWebContext();
 
   const {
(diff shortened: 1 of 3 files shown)


frontend/packages/client/src/utils.js
```diff
@@ -184,3 +184,10 @@ export const customDraftToHTML = (content) => {
 };
 
 export const isValidAddress = (addr) => /0[x,X][a-zA-Z0-9]{16}$/gim.test(addr);
+
+export const wait = async (milliSeconds = 5000) =>
+  new Promise((resolve) => {
+    setTimeout(() => {
+      resolve(true);
+    }, milliSeconds);
+  });
```
germanurrus, line 193: this is not used anywhere in the production code, but it's used when trying to create a delay on loading states.

frontend/packages/client/src/pages/CommunityCreate.js
```diff
@@ -65,13 +66,70 @@ export default function CommunityCreate() {
...
+      listAddrAuthors = [],
+      strategies = [],
+    } = fields;
+
+    // validate Flow Addresses used
+    const addressesToValidate = {
+      'Contract Address': [contractAddress],
+      'Admin List': listAddrAdmins.map((e) => e.addr),
+      'Author List': listAddrAuthors.map((e) => e.addr),
+      Strategies: strategies,
```
germanurrus, line 81: Strategies will be added on https://github.com/DapperCollectives/CAST/pull/62, adding them here to update Pr after #62 is merged

dbslone (approved):

PR #84, opened 2022-06-28
CAS-139: Add EarlyVote scoring

jbluks:
**Ticket**: [CAS-139](https://linear.app/dappercollectives/issue/CAS-139/add-early-voting-points-to-leaderboard-algorithm)

**Description:**

- Adds table `community_users_achievements` which will track user achievements
- Adds logic to add row to `community_users_achievements` when user does an early vote
- Adds logic to `/communities/:communityId/leaderboard` to add early vote bonuses to user score


backend/main/models/vote.go
```diff
@@ -194,13 +200,36 @@ func (v *Vote) GetVoteById(db *s.Database) error {
...
 	`, v.Proposal_id, v.Addr, v.Choice, v.Composite_signatures, v.Cid, v.Message).Scan(&v.ID, &v.Created_at)
 
+	// Get Proposal Start time to check for early vote
+	var proposal Proposal
+	pgxscan.Get(db.Context, db.Conn, &proposal,
+		`SELECT start_time, community_id from proposals
+		WHERE id = $1`,
+		v.Proposal_id)
+
+	isEarlyVote := proposal.Start_time.Before(proposal.Start_time.Add(time.Hour + time.Duration(defaultEarlyVoteLength)))
```
dbslone, line 220: Maybe i'm reading this wrong but i think this would always evaluate as true. Should this compare if the current time is less than 1 hour from the proposal start time?

backend/main/leaderboard_test.go
```diff
@@ -34,4 +36,33 @@ func TestGetCommunityLeaderboard(t *testing.T) {
 	assert.Equal(t, 2, len(p.Data))
 	assert.Equal(t, 3, p.Data[0].Score)
 	assert.Equal(t, 2, p.Data[1].Score)
+}
+
+func TestGetCommunityLeaderboardWithEarlyVotes(t *testing.T) {
+	clearTable("communities")
+	clearTable("community_users")
+	clearTable("proposals")
+	clearTable("votes")
```
dbslone, line 45: should the `community_users_achievements` be cleared before this test runs?

backend/main/models/community_users.go
```diff
@@ -98,29 +100,54 @@ func GetUsersForCommunityByType(db *s.Database, communityId, start, count int, u
...
 		return nil, 0, err
 	} else if err != nil && err.Error() == pgx.ErrNoRows.Error() {
-		return []LeaderboardUser{}, 0, nil
+		return []LeaderboardUserPayload{}, 0, nil
+	}
+
+	for _, user := range userAchievements {
+		var leaderboardUser = LeaderboardUserPayload{}
+		leaderboardUser.Addr = user.Address
+		leaderboardUser.Score = user.NumVotes + (user.EarlyVote * defaultEarlyVoteWeight)
```
dbslone, line 142: Is the idea here that when we enable user editable weights that this value will not be 1? If so maybe it's better to add `earlyVoteWeight` as an argument to the method and assign a default value if it is not passed?

backend/main/models/vote.go
jbluks: Yep - good catch. That is supposed to be checking `v.Created_at` - which is when the user voted and comparing if that is before the early vote cut off on the proposal.

backend/main/leaderboard_test.go
jbluks: Good catch.

backend/migrations/000018_add_achievements_table.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS community_users_achievements;
```
jacksonConrad, line 1: Can you also `DROP TYPE IF EXISTS achievement_types` here?

backend/main/models/vote.go
jacksonConrad, line 220: should this be a product instead of a sum?

i.e. `time.Hour * time.Duration(defaultEarlyVoteLength)`?

backend/main/models/community_users.go
```diff
@@ -98,29 +100,54 @@ func GetUsersForCommunityByType(db *s.Database, communityId, start, count int, u
...
-		GROUP BY v.addr
-		ORDER BY score DESC
-		LIMIT $2 OFFSET $3
-		`, communityId, count, start)
+		SELECT v.addr as address, count(*) as num_votes, 
+			CASE WHEN a.early_vote is NULL THEN 0 ELSE a.early_vote END as early_vote 
+			FROM votes v 
+			LEFT OUTER JOIN proposals p ON p.id = v.proposal_id
+			LEFT OUTER JOIN (
+				SELECT * FROM crosstab(
```
jacksonConrad, line 119: This SQL is breaking my brain 😄 

Can you leave a comment roughly describing what is going on here?

jacksonConrad (changes requested):

backend/migrations/000018_add_achievements_table.down.sql
jbluks: Done.

backend/main/models/vote.go
jbluks: Yep - good catch, I wasn't understanding properly how the `time.Duration` was working.

backend/main/models/community_users.go
jbluks: Done

rvmelkonian:
I'm getting a failing test when I run `make test`
```
TestGetCommunityLeaderboard (0.03s)
panic: runtime error: index out of range [0] with length 0 [recovered]
        panic: runtime error: index out of range [0] with length 0
```

backend/main/models/community_users.go
rvmelkonian: yeah.. Go doesn't support optional fields, Golang enums can be useful if you need to pass different types into the same func tho!

For default values, you could use the constructor pattern, so where leaderboardUserPayload would have a function to initialise it with a default value.

jbluks: Yea, once get it all working, we can revisit when we bring in the custom settings support.

backend/main/models/community_users.go
```diff
@@ -98,29 +100,58 @@ func GetUsersForCommunityByType(db *s.Database, communityId, start, count int, u
...
+		leaderboardUser.Score = user.NumVotes + (user.EarlyVote * defaultEarlyVoteWeight)
+		leaderboardUsers = append(leaderboardUsers, leaderboardUser)
 	}
 
 	var totalUsers int
 	countSql := `SELECT COUNT(*) FROM community_users WHERE community_id = $1`
 	_ = db.Conn.QueryRow(db.Context, countSql, communityId).Scan(&totalUsers)
 
-	return users, totalUsers, nil
+	return leaderboardUsers, totalUsers, nil
```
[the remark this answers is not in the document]
jbluks: Updated to use helper functions.

backend/main/leaderboard_test.go
```diff
@@ -34,4 +38,33 @@ func TestGetCommunityLeaderboard(t *testing.T) {
...
+	communityId := otu.AddCommunities(1)[0]
+	proposalIds := otu.AddActiveProposalsWithStartTimeNow(communityId, 3)
+	voteChoice := "a"
+	earlyVoteBonus := 1
+
+	otu.CreateVoteAPI(proposalIds[0], otu.GenerateValidVotePayload("user1", proposalIds[0], voteChoice))
+	otu.CreateVoteAPI(proposalIds[1], otu.GenerateValidVotePayload("user1", proposalIds[1], voteChoice))
+	otu.CreateVoteAPI(proposalIds[2], otu.GenerateValidVotePayload("user1", proposalIds[2], voteChoice))
+	otu.CreateVoteAPI(proposalIds[0], otu.GenerateValidVotePayload("user2", proposalIds[0], voteChoice))
+	otu.CreateVoteAPI(proposalIds[1], otu.GenerateValidVotePayload("user2", proposalIds[1], voteChoice))
```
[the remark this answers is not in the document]
jbluks: Done.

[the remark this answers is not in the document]
rvmelkonian: sweet! looks clean

rvmelkonian (approved):

PR #95, opened 2022-07-04
intergrate snapshot server to cast API WIP

rvmelkonian:


backend/main/models/proposal.go
```diff
@@ -35,6 +35,7 @@ type Proposal struct {
 	Timestamp            string                  `json:"timestamp" validate:"required"`
 	Composite_signatures *[]s.CompositeSignature `json:"compositeSignatures" validate:"required"`
 	Computed_status      *string                 `json:"computedStatus,omitempty"`
+	Snapshot             s.LatestBlockHeight     `json:"snapshot,omitempty"`
```
jacksonConrad, line 38: i think this should be a pointer, if there is a possibility of the field being empty

backend/main/server/app.go
```diff
@@ -709,6 +707,26 @@ func (a *App) getProposal(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	strategy, err := models.MatchStrategyByProposal(*c.Strategies, *p.Strategy)
+	if err != nil {
+		respondWithError(w, http.StatusInternalServerError, err.Error())
+		return
+	}
+
+	snapshotResponse, err := a.SnapshotClient.GetSnapshotStatus(strategy.Contract)
```
jacksonConrad, line 722: we might want to make this request conditional depending on the strategy, since only a few of the strategies require the snapshotter

backend/main/server/app.go
```diff
@@ -799,7 +817,21 @@ func (a *App) createProposal(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	// pin to ipfs
+	if err := a.SnapshotClient.TakeSnapshot(strategy.Contract); err != nil {
```
jacksonConrad, line 820: same here, we should only trigger snapshots for strategies that require it

jacksonConrad (changes requested):
looking great! just a few little changes

jacksonConrad (changes requested):
one last thing :)

backend/main/server/app.go
```diff
@@ -752,8 +775,7 @@ func (a *App) createProposal(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// get latest snapshotted blockheight
-	snapshot, err := a.SnapshotClient.GetLatestSnapshot()
+	snapshot, err := a.SnapshotClient.GetLatestFlowSnapshot()
```
jacksonConrad, line 778: This call is here to set `p.Block_height`, which works for any FlowToken based strategy, but now that we are triggering a snapshot in certain cases, we need to use the block height that the triggered snapshot is going to be for.

So maybe we do need the `GetLatestSnapshot` call down below, or if you can find a way to return the blockheight being snapshotted from the `TakeSnapshot` API call, that would be the best solution with the least number of network requests.

jacksonConrad (changes requested):

rvmelkonian:
Cool! @jacksonConrad those updates are done, and dependant on this PR for the snapshotrer https://github.com/brudfyi/snapshot-flow-balances/pull/10

backend/main/server/app.go
```diff
@@ -719,19 +719,16 @@ func (a *App) getProposal(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var snapshotResponse shared.LatestBlockHeight
-
-	//check if strategy has a Contract field
 	if strategy.Contract.Name != nil {
-		snapshotResponse, err = a.SnapshotClient.GetSnapshotStatus(strategy.Contract)
+		snapshotResponse, err := a.SnapshotClient.GetSnapshotStatus(strategy.Contract)
```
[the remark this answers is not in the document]
dbslone: @jacksonConrad I think the first approach will be good for now keeps the FE from needing to make multiple requests

backend/main/server/app.go
```diff
@@ -1049,17 +1034,17 @@ func (a *App) createCommunity(w http.ResponseWriter, r *http.Request) {
...
-		if err != nil {
-			respondWithError(w, http.StatusInternalServerError, "IPFS error: "+err.Error())
-			return
-		}
-		c.Cid = &pin.IpfsHash
-	} else {
-		dummyCid := "000000000000000"
-		c.Cid = &dummyCid
-	}
+	// if os.Getenv("APP_ENV") != "TEST" {
```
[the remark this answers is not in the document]
rvmelkonian: was for local testing, when `APP_ENV == "DEV"` will add back in. Turns out my wifi booster was the problem, when not connected to it, I'm able to pin to ipfs.

backend/main/server/app.go
```diff
@@ -817,20 +822,22 @@ func (a *App) createProposal(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
 
-	snapshotResponse, err := a.SnapshotClient.GetSnapshotStatus(strategy.Contract)
-	if err != nil {
-		log.Error().Err(err).Msg("error getting snapshot status")
-		respondWithError(w, http.StatusInternalServerError, err.Error())
-		return
-	}
+		snapshotResponse, err := a.SnapshotClient.GetSnapshotStatus(strategy.Contract)
```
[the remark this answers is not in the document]
rvmelkonian: this has been implemented now

rvmelkonian:
this is ready to be reviewed again. This PR also merges in all the changes from cas 217.

backend/main/server/app.go
```diff
@@ -1875,3 +1865,21 @@ func (a *App) validateTimestamp(timestamp string, expiry int) error {
...
+func (a *App) processSnapshotStatus(s *models.Strategy, p *models.Proposal) error {
+	var processing = "processing"
+	if s.Contract.Name != nil && p.Snapshot_status == &processing {
+		snapshotResponse, err := a.SnapshotClient.GetSnapshotStatus(s.Contract)
+		if err != nil {
+			return err
+		}
+
+		p.Snapshot_status = &snapshotResponse.Data.Status
+		p.Block_height = snapshotResponse.Data.BlockHeight
```
jacksonConrad, line 1878: blockheight doesnt need to be added, is already on proposal

backend/main/server/app.go
```diff
@@ -1875,3 +1865,21 @@ func (a *App) validateTimestamp(timestamp string, expiry int) error {
 	}
 	return nil
 }
+
+func (a *App) processSnapshotStatus(s *models.Strategy, p *models.Proposal) error {
+	var processing = "processing"
+	if s.Contract.Name != nil && p.Snapshot_status == &processing {
+		snapshotResponse, err := a.SnapshotClient.GetSnapshotStatus(s.Contract)
```
jacksonConrad, line 1872: Get snapshot at specific blockheight

backend/main/server/app.go
```diff
@@ -756,36 +767,38 @@ func (a *App) createProposal(w http.ResponseWriter, r *http.Request) {
...
 	}
-
 	strategy, err := models.MatchStrategyByProposal(*community.Strategies, *p.Strategy)
 	if err != nil {
 		log.Error().Err(err).Msg("Community does not have this strategy availabe")
 		respondWithError(w, http.StatusInternalServerError, err.Error())
 		return
+
+	}
+	snapshot, err := a.SnapshotClient.GetLatestSnapshot(strategy.Contract)
```
jacksonConrad, line 792: Replace this block w./ TakeSnapshot

backend/main/server/app.go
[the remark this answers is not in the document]
jacksonConrad: processSnapshotStatus here instead of GetSnapshotStatus

jacksonConrad (changes requested):

rvmelkonian:
ok I believe all is done! the updates are dependant on a new PR to the snapshotter https://github.com/brudfyi/snapshot-flow-balances/pull/13

backend/main/server/app.go
rvmelkonian: removed

backend/main/server/app.go
rvmelkonian: renamed to `GetSnapshotStatusAtBlockHeight` and updated logic. Room for improvement in `setSnapshotUrl` but I have made a separate ticket for it

backend/main/server/app.go
rvmelkonian: pesky merge conflict. Reverted back to its superior form ...

backend/main/server/app.go
```diff
@@ -784,8 +783,19 @@ func (a *App) createProposal(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	snapshotResponse, err := a.SnapshotClient.TakeSnapshot(strategy.Contract)
+	if err != nil {
+		log.Error().Err(err).Msg("error taking snapshot")
+		respondWithError(w, http.StatusInternalServerError, err.Error())
+		return
+	}
+
+	p.Snapshot_status = &snapshotResponse.Data.Status
```
jacksonConrad, line 796: also need to set `p.Block_height here`

rvmelkonian: on it!

rvmelkonian: that's done and pushed

jacksonConrad (approved):
looks good!

PR #96, opened 2022-07-04
CAS-170 - validation on Proposal & Voting step of community creation

germanurrus:
To test this PR go to `http://localhost:3000/#/community/create` and get to the Proposal & Voting Step
Validation was changed so that:
- Path 1: Setting a threshold
User fills out all 4 fields
Create button becomes active
User clicks to proceed
(authors checkbox could or could not be checked)
Screen Shot 2022-07-04 at 17 44 02

- Path 2: Selecting the Authors-Only checkbox
User clicks the authors-only checkbox
Create button becomes active
User clicks to proceed
Screen Shot 2022-07-04 at 17 43 22


frontend/packages/client/src/hooks/useCommunity.js
```diff
@@ -106,17 +112,20 @@ export default function useCommunity() {
...
-            contractAddress,
-            contractName,
-            storagePath,
+            contractAddress: setDefaultValue(
+              contractAddr,
+              '0x0ae53cb6e3f42a79'
+            ),
+            contractName: setDefaultValue(contractN, 'FlowToken'),
+            storagePath: setDefaultValue(storageP, 'flowTokenBalance'),
+            proposalThreshold: setDefaultValue(propThreshold, '0'),
```
germanurrus, line 128: as discussed with @rvmelkonian if this fields are not filled in the from, frontend needs to send these values as default

dbslone (approved):

PR #98, opened 2022-07-05
CAS-140: Add Streak Bonus

jbluks:
**Ticket:** [CAS-140](https://linear.app/dappercollectives/issue/CAS-140/add-streak-points-to-leaderboard-algorithm)

**Description**

- Adds streak check to vote that adds achievements to user for vote streaks
- Refactors code to put queries into helper functions


backend/main/leaderboard_test.go
```diff
@@ -34,4 +32,71 @@ func TestGetCommunityLeaderboard(t *testing.T) {
 	assert.Equal(t, 2, len(p.Data))
 	assert.Equal(t, 3, p.Data[0].Score)
 	assert.Equal(t, 2, p.Data[1].Score)
```
rvmelkonian, line 34: I wonder if we can be clear about what is being asserted here by storing these values in named variables?

backend/main/test_utils/leaderboard_utils.go
```diff
@@ -0,0 +1,52 @@
...
+func (otu *OverflowTestUtils) GenerateLeaderboardBaseCase(communityId int) {
+	proposalIds := otu.AddActiveProposals(communityId, 3)
+	voteChoice := "a"
+
+	otu.CreateVoteAPI(proposalIds[0], otu.GenerateValidVotePayload("user1", proposalIds[0], voteChoice))
+	otu.CreateVoteAPI(proposalIds[1], otu.GenerateValidVotePayload("user1", proposalIds[1], voteChoice))
+	otu.CreateVoteAPI(proposalIds[2], otu.GenerateValidVotePayload("user1", proposalIds[2], voteChoice))
+	otu.CreateVoteAPI(proposalIds[0], otu.GenerateValidVotePayload("user2", proposalIds[0], voteChoice))
+	otu.CreateVoteAPI(proposalIds[1], otu.GenerateValidVotePayload("user2", proposalIds[1], voteChoice))
+}
```
rvmelkonian, line 12: for this function and for `GenerateLeaderboardWithEarlyVotes` you could pass in a parameter of type `proposalIds[]` iterate over it and get the values at each index to pass in to `CreateVoteApi`

backend/main/leaderboard_test.go
```diff
@@ -34,4 +32,71 @@ func TestGetCommunityLeaderboard(t *testing.T) {
...
+
+	assert.Equal(t, 2, len(p.Data))
+	assert.Equal(t, 2+2*earlyVoteBonus, p.Data[0].Score)
+	assert.Equal(t, 2, p.Data[1].Score)
+}
+
+func TestGetCommunityLeaderboardWithSingleStreak(t *testing.T) {
+	clearTable("communities")
+	clearTable("community_users")
+	clearTable("community_users_achievements")
```
rvmelkonian, line 63: just a small thing, but maybe we can call this table `achievements`? to avoid the long name?

backend/main/models/vote.go
```diff
@@ -194,14 +201,27 @@ func (v *Vote) GetVoteById(db *s.Database) error {
...
+	}
+
+	proposal, err := getProposal(db, v.Proposal_id)
+
+	if checkError(err) {
+		return err
+	}
+
+	err = checkForEarlyVoteAchievement(db, v, proposal)
+
```
rvmelkonian, line 217: could you remove these blank lines before the `checkError`? So that we handle the error directly after the function call with no whitespace.

backend/main/models/vote.go
```diff
@@ -308,3 +328,145 @@ func DoesNFTExist(db *s.Database, v *VoteWithBalance) (bool, error) {
...
+					sql = `
+						INSERT INTO community_users_achievements(addr, achievement_type, community_id, proposals, details)
+						VALUES($1, $2, $3, $4, $5)
+						ON CONFLICT (details)
+						DO UPDATE SET proposals = $4, details = $6
+						RETURNING id
+					`
+					err = db.Conn.QueryRow(db.Context, sql, v.Addr, Streak, p.Community_id, proposals, previousStreakDetails, currentStreakDetails).Scan(&v.ID)
+				}
+
```
rvmelkonian, line 442: I'm struggling to comprehend this block of code in the for loop. I think it's because of the amount of logic being exposed to the reader at one time. Maybe breaking it up will make it easier to read? So before you go into the for loop that could be done in a separate function with a meaningful name.

Also here
`// Add default streak, or do nothing if streak already exists`. Just include it in a function with a name like `AddDefaultStreak` then we might not need to add a comment, as the function name and parameters will be self-explanatory.

also ``// If previous streak exists, then update with new streak details`` new function call `updateStreakDetails` it will be easier to comprehend this way I think, and means each function is just doing one part of the logic. Will save other developers if they need to debug something.

backend/main/models/vote.go
```diff
@@ -308,3 +328,145 @@ func DoesNFTExist(db *s.Database, v *VoteWithBalance) (bool, error) {
...
+	var proposal Proposal
+	err := pgxscan.Get(db.Context, db.Conn, &proposal,
+		`SELECT start_time, community_id from proposals
+		WHERE id = $1`,
+		proposalId)
+
+	return proposal, err
+}
+
+func checkForEarlyVoteAchievement(db *s.Database, v *Vote, p Proposal) error {
```
rvmelkonian, line 354: it's not clear from the function name that this function actually does an `INSERT` and updates the DB. This function is doing two things, checking `isEarlyVote` then it's also updating the DB.  

I think we should put the sql part into a separate function called `updateUserAchievments` then, at the controller level conditionally call that if isEarlyVote is true, which would be returned from this function.

rvmelkonian:
few merge conflicts here

backend/main/test_utils/leaderboard_utils.go
jbluks: Done.

backend/main/leaderboard_test.go
jbluks: Done.

backend/main/models/vote.go
jbluks: Done.

rvmelkonian:
just added one last suggestion, other than that looks good!

rvmelkonian (approved):

backend/main/models/vote.go
```diff
@@ -202,35 +208,27 @@ func (v *Vote) GetVoteById(db *s.Database) error {
...
-	err := db.Conn.QueryRow(db.Context,
-		`
-		INSERT INTO votes(proposal_id, addr, choice, composite_signatures, cid, message)
-		VALUES($1, $2, $3, $4, $5, $6)
-		RETURNING id, created_at
-	`, v.Proposal_id, v.Addr, v.Choice, v.Composite_signatures, v.Cid, v.Message).Scan(&v.ID, &v.Created_at)
+	err := createVote(db, v)
+	if checkError(err) {
+		return err
+	}
```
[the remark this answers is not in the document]
jbluks: Done.

PR #97, opened 2022-07-05
CAS-154 - add My Communities section to homepage

germanurrus:
Home page updates:

Screen Shot 2022-07-05 at 17 45 11

- Added `react-query`
- Remove redux dependencies not used 
- Moved some `dependencies` to `dev dependencies` 

There's only one issue that will be addressed in another PR:
 1. User clicks on join in a community in the home page from the feature communities section
 2. If the "My Communities" section was not visible it should appear with the community the user has just joined. Same if the user leaves the community and that was the only one, the "My Communities"  section should fade out

PR #108, opened 2022-07-07
CAS-174 - update my communities section after joining a community

germanurrus:
Depends on https://github.com/DapperCollectives/CAST/pull/97
- Use mutations for adding and removing users from communities. 
- Updates home screen when user is added as member

### DIFF
--- a/frontend/packages/client/src/App.js
+++ b/frontend/packages/client/src/App.js
@@ -15,7 +15,7 @@ import { QueryClient, QueryClientProvider } from 'react-query';
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
-      refetchOnWindowFocus: false,
+      refetchOnWindowFocus: process.env.NODE_ENV !== 'development',
     },
   },
 });
--- a/frontend/packages/client/src/App.js
+++ b/frontend/packages/client/src/App.js
@@ -9,6 +9,16 @@ import { ErrorHandler } from './components';
 import NotificationModalProvider from './contexts/NotificationModal';
 import { ErrorBoundary } from 'react-error-boundary';
 import AppPages from './pages';
+import { QueryClient, QueryClientProvider } from 'react-query';
+
+// create react-query client
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: false,
+    },
+  },
+});
 
 function App() {
   return (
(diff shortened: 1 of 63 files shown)


frontend/packages/client/src/App.js
```diff
@@ -18,15 +30,18 @@ function App() {
...
+        <Web3Provider network={process.env.REACT_APP_FLOW_ENV}>
+          <NotificationModalProvider>
+            <Router>
+              <ErrorHandler>
+                <AppPages />
+              </ErrorHandler>
+            </Router>
+          </NotificationModalProvider>
+        </Web3Provider>
+        {isLocalDevMode && <ReactQueryDevtools />}
```
germanurrus, line 43: This enables the react-query debugger: 
Screen Shot 2022-07-07 at 17 34 57

frontend/packages/client/src/hooks/useJoinCommunity.js
```diff
@@ -1,74 +1,77 @@
...
-      if (_compositeSignatures.indexOf('Declined:') > -1) {
-        dispatch({ type: 'SUCCESS' });
-        return { success: false };
-      }
-      const compositeSignatures = getCompositeSigs(_compositeSignatures);
-      if (!compositeSignatures) {
-        return { error: 'No valid user signature found.' };
-      }
+  const { mutateAsync: createCommunityUserMutation, mutate: mutateOne } =
+    useMutation(
```
germanurrus, line 10: I had tu use `mutateAsync` as opposed to `mutate` because of the way the hook is used in the component. The component is expecting to get a result from the mutation like here: 

In the docs: https://react-query.tanstack.com/reference/useMutation  retuning the mutate function seems the way to go, they we could use something like:
```
const { mutate, data, error, isLoading } = useSomeWrapperFunctionMutation(mutate, 
{
   onSuccess: () => {
        manage side effects here
   }
   });
then we can call
mutate(payload);

or if we have specific side effects:
mutate(payload, {
   onSuccess: (data, variables, context) => {
     // I will fire second!
   },
   onError: (error, variables, context) => {
     // I will fire second!
   },
   onSettled: (data, error, variables, context) => {
     // I will fire second!
   },
 })

File: frontend/packages/client/src/hooks/useJoinCommunity.js [hunk 3]
```diff
@@ -1,27 +1,21 @@
...
-  const [state, dispatch] = useReducer(defaultReducer, {
-    ...INITIAL_STATE,
-    loading: false,
-  });
+  const queryClient = useQueryClient();
   const { notifyError } = useErrorHandlerContext();
 
-  const createCommunityUser = useCallback(
-    async (communityId, user, injectedProvider) => {
+  const { mutateAsync: createCommunityUserMutation } = useMutation(
```
germanurrus, line 9: I had tu use `mutateAsync` as opposed to `mutate` because of the way the hook is used in the component. The component is expecting to get a result from the mutation like here: https://github.com/DapperCollectives/CAST/pull/108/files#diff-ba7b9f47b489056dd6b487fee2e4bf14d1d836b8f8de40956a83cd57e1e84b39R34

In the docs: https://react-query.tanstack.com/reference/useMutation  retuning the mutate function seems the way to go, they we could use something like:
```
const { mutate, data, error, isLoading } = useSomeWrapperFunctionMutation(mutate, 
{
   onSuccess: () => {
        manage side effects here; general ones like some query invalidation adding data to loaded queries and so on
   }
});
   
then we can call
mutate(payload);

or if we have specific side effects:
mutate(payload, {
   onSuccess: (data, variables, context) => {
     // something we need to happen on the component that triggered the update
   },
   onError: (error, variables, context) => {
     // same here
   },
   onSettled: (data, error, variables, context) => {
     // same here
   },
 })

jbluks (approved):
LGTM

dbslone (approved):

PR #115, opened 2022-07-08
cas 202 update-backend-endpoint

rvmelkonian:
bugfix for dup communities if user has multiple roles

* adds `getUniqueCommunities` function

### DIFF
--- a/backend/main/models/community_users.go
+++ b/backend/main/models/community_users.go
@@ -162,7 +162,20 @@ func GetCommunitiesForUser(db *s.Database, addr string, start, count int) ([]Use
 	`
 	_ = db.Conn.QueryRow(db.Context, countSql, addr).Scan(&totalCommunities)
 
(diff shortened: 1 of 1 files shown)


germanurrus:
hey @rvmelkonian I did a test on this endpoint `/users/{addr:0x[a-zA-Z0-9]{16}}/communities` is it possible to return in the field `membershipType: "member"` something like an array `membershipType: ["member", "author", "admin"]` ?The reason is because we are using this endpoint in the FE to check for user roles within a community or to know a list of communities where an address has access. 
Now after this change if you create a community it's returning `membershipType: "member"` and FE can't termine if it's an admin for showing the edit community button.

rvmelkonian:
gotya. let me have a fiddle

dbslone:
@rvmelkonian you can close this PR in favor of fixing how we manage roles. For now this is resolved on the frontend

PR #125, opened 2022-07-11
CAS-188 - show join button when user is not connected

germanurrus:
Updates to show "Join" button
Screen Shot 2022-07-11 at 11 58 18

Screen Shot 2022-07-11 at 11 59 07


frontend/packages/client/src/components/Community/JoinCommunityButton.js
```diff
@@ -10,18 +10,12 @@ export default function JoinCommunityButton({
 }) {
   const { createCommunityUser, deleteUserFromCommunity } = useJoinCommunity();
   const { injectedProvider, user } = useWebContext();
-  const [addr, setAddr] = useState();
-  const memberState = useUserRoleOnCommunity({
-    addr,
+
+  const isMember = useUserRoleOnCommunity({
+    addr: user?.addr,
```
germanurrus, line 15: using hook output as input on another hook so we don't need create the `addr` state

dbslone (approved):

PR #133, opened 2022-07-12
CAS-177: Integrate Leaderboard API with front-end

jbluks:
**Ticket:** [CAS-177](https://linear.app/dappercollectives/issue/CAS-177/add-leaderboard-score-to-community-member-page)

**Description**

Connects Leaderboard API with the Leaderboard component on the community page

Screen Shot 2022-07-12 at 11 16 41 AM

NOTE: Added a pattern to separate the API logic from the useQuery hook. This promotes reusability of the api logic, and simplifies the fetching hooks so that the only logic relates to the hook, and not to the fetching.

### DIFF
--- a/frontend/packages/client/src/api/constants.js
+++ b/frontend/packages/client/src/api/constants.js
@@ -1,3 +1,3 @@
 export const API_BASE_URL = process.env.REACT_APP_BACK_END_SERVER_API;
 
-export const COMMUNITIES_URL = `${API_BASE_URL}/communities/`;
+export const COMMUNITIES_URL = `${API_BASE_URL}/communities`;
--- a/frontend/packages/client/src/api/constants.js
+++ b/frontend/packages/client/src/api/constants.js
@@ -1,0 +1,3 @@
+export const API_BASE_URL = process.env.REACT_APP_BACK_END_SERVER_API;
+
+export const COMMUNITIES_URL = `${API_BASE_URL}/communities/`;
--- a/frontend/packages/client/src/api/leaderboard.js
+++ b/frontend/packages/client/src/api/leaderboard.js
@@ -1,0 +1,17 @@
+import { checkResponse } from '../utils';
+import { COMMUNITIES_URL } from './constants';
+
+const LEADERBOARD_SIZE = 10;
+const getLeaderBoardUrl = (communityId) =>
+  `${COMMUNITIES_URL}/${communityId}/leaderboard?count=${LEADERBOARD_SIZE}`;
+
+const fetchLeaderBoard = async (communityId) => {
+  try {
+    const response = await fetch(getLeaderBoardUrl(communityId));
+    return await checkResponse(response);
+  } catch (err) {
+    throw err;
+  }
+};
+
+export default fetchLeaderBoard;
--- a/frontend/packages/client/src/api/leaderboard.js
+++ b/frontend/packages/client/src/api/leaderboard.js
@@ -1,13 +1,13 @@
 import { checkResponse } from '../utils';
 import { COMMUNITIES_URL } from './constants';
 
-const LEADERBOARD_SIZE = 10;
-const getLeaderBoardUrl = (communityId) =>
-  `${COMMUNITIES_URL}/${communityId}/leaderboard?count=${LEADERBOARD_SIZE}`;
+const DEFAULT_PAGE_SIZE = 10;
+const getLeaderBoardUrl = (communityId, pageSize) =>
+  `${COMMUNITIES_URL}/${communityId}/leaderboard?count=${pageSize}`;
 
-const fetchLeaderBoard = async (communityId) => {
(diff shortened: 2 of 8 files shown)


frontend/packages/client/src/api/leaderboard.js
```diff
@@ -0,0 +1,17 @@
...
+  `${COMMUNITIES_URL}/${communityId}/leaderboard?count=${LEADERBOARD_SIZE}`;
+
+const fetchLeaderBoard = async (communityId) => {
+  try {
+    const response = await fetch(getLeaderBoardUrl(communityId));
+    return await checkResponse(response);
+  } catch (err) {
+    throw err;
+  }
+};
```
germanurrus, line 15: I like this approach of adding this in a separate file!

What do you think about making a variable `pageSize` a param to the Url then if that changes on the hook it's already prepared

```suggestion
const LEADERBOARD_SIZE = 10;
const getLeaderBoardUrl = (communityId, pageSize) =>
  `${COMMUNITIES_URL}/${communityId}/leaderboard?count=${pageSize}`;

const fetchLeaderBoard = async (communityId, pageSize = LEADERBOARD_SIZE ) => {
  try {
    const response = await fetch(getLeaderBoardUrl(communityId, pageSize));
    return await checkResponse(response);
  } catch (err) {
    throw err;
  }
};
```

frontend/packages/client/src/api/leaderboard.js
```diff
@@ -0,0 +1,17 @@
+import { checkResponse } from '../utils';
+import { COMMUNITIES_URL } from './constants';
```
germanurrus, line 2: looks like there's a double `/` . Here and then on line 6 it's adding another `/` Causing a fetch error.

frontend/packages/client/src/api/leaderboard.js
jbluks: Oh yes - that is a better approach.

frontend/packages/client/src/api/leaderboard.js
jbluks: Fixed.

frontend/packages/client/src/api/constants.js
```diff
@@ -0,0 +1,3 @@
+export const API_BASE_URL = process.env.REACT_APP_BACK_END_SERVER_API;
+
+export const COMMUNITIES_URL = `${API_BASE_URL}/communities/`;
```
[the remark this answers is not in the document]
jbluks: I added this as a potential new pattern, so we don't have process.env.REACT_APP... all over the app. The idea is that base URLS are built in here and used by any API call that needs them. Going forward, all api fetching/updating will be defined in this directory and use these constants. For example the COMMUNITIES_URL will be used by all queries to that end point.

frontend/packages/client/src/hooks/useLeaderBoard.js
```diff
@@ -1,70 +1,29 @@
...
+    isError,
+    error,
+    data: {
+      leaderBoard: data?.data ?? [],
+      currentUser: {
+        addr: 'joshprin...',
+        score: '123 $XYZ',
+        index: 122,
+      },
+    },
```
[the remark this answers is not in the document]
jbluks: Yea - I will follow up with Product on this. If we make a change, will create a new ticket for it.

germanurrus (commented):
Looks good 👍 (FE side) , thanks for the updates! just left some comments

frontend/packages/client/src/hooks/useLeaderBoard.js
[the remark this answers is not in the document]
jbluks: Followed up with Jose and it is expected that we show the current user and their score. I have created a ticket to add this feature: [CAS-129](https://linear.app/dappercollectives/issue/CAS-229/add-current-user-to-leaderboard)

frontend/packages/client/src/pages/Community.js
```diff
@@ -185,7 +185,7 @@ export default function Community() {
 
   // these two fields should be coming from backend as configuration
   const showPulse = false;
-  const showLeaderBoard = false;
+  const showLeaderBoard = true;
```
CJamesKeller, line 188: Do we have a ticket to grab this from the backend? If not, let's add one.

CJamesKeller (approved):

frontend/packages/client/src/components/LeaderBoard.js
```diff
@@ -74,7 +77,7 @@ export default function LeaderBoard({ onClickViewMore = () => {} } = {}) {
             })}
```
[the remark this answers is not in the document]
germanurrus: can you add the classes: `is-flex flex-1 is-justify-content-center has-text-weight-bold pr-2` to make it centered, it's looking like this:
Screen Shot 2022-07-14 at 12 32 21

germanurrus (approved):
Looks good, thanks @jbluks for the changes!

rvmelkonian (approved):
looks good, nice with the tests!

PR #135, opened 2022-07-12
Rich/cas 217 fix snapshot balance vote in `createVoteForProposal` controller

rvmelkonian:
This is a subticket of cas-216. Some logic was missing in the Strategy `FetchBalance` method. Also now dummy balance data is returned for local dev.

### DIFF
--- a/backend/main/server/app.go
+++ b/backend/main/server/app.go
@@ -643,9 +643,6 @@ func (a *App) createVoteForProposal(w http.ResponseWriter, r *http.Request) {
 		v.Cid = &dummyCid
 	}
 
-	//log the vote.Choice
-	log.Info().Msgf("Vote : %+v", v)
-
 	if err := v.CreateVote(a.DB); err != nil {
 		log.Error().Err(err).Msg("Couldnt create vote")
 		respondWithError(w, http.StatusInternalServerError, err.Error())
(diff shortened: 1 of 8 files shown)


dbslone:
We will forget in the future why there is fragmentation in the code. Can you write a quick doc in `docs/<filename>.md` that explains how this operates? Doesn't need to be detailed just need a quick high level explanation for now.

germanurrus:
I tested locally with FE creating and voting a proposal with `token-weighted-default` and works fine

backend/main/shared/snapshot-flow.go
```diff
@@ -23,13 +23,35 @@ type Snapshot struct {
 	Finished     time.Time `json:"finished"`
 }
 
+// copied this quick and dirty from models to get around circular dependency
+// move to shared?
+type Balance struct {
```
jacksonConrad, line 28: Yeah probably cleanest to move this to `shared/structs.go` and use it in `balance.go` and `snapshot-flow.go` if that works

backend/main/server/app.go
```diff
@@ -547,11 +547,13 @@ func (a *App) createVoteForProposal(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// check that proposal is live
-	if !p.IsLive() {
-		err = errors.New("user cannot vote on inactive proposal")
-		log.Error().Err(err)
-		respondWithError(w, http.StatusInternalServerError, err.Error())
-		return
+	if os.Getenv("APP_ENV") != "DEV" {
```
jacksonConrad, line 550: is this just to make testing locally easier?

jacksonConrad (changes requested):

rvmelkonian:
clsoing this as it was merged into `cas-172` which has now been merged to main.

PR #138, opened 2022-07-12
cas-107 Delay proposal start by 1 hour in production env (UI ONLY)

CJamesKeller:
[cas-107](https://linear.app/dappercollectives/issue/CAS-107/increase-time-before-proposals-can-start)


frontend/packages/client/src/components/ProposalCreate/StepTwo.js
```diff
@@ -47,7 +47,10 @@ const StepTwo = ({ stepData, setStepValid, onDataChange }) => {
 
     // push now if date is today and not already in time interval
     if (cutOffDate) {
-      const nowDate = new Date();
+      const nowDate =
+        process.env.NODE_ENV === 'production'
```
dbslone, line 51: Need to introduce a new environment variable like `APP_ENV` because staging uses `NODE_ENV=production`

https://github.com/DapperCollectives/CAST/blob/main/frontend/Dockerfile.stage#L11

dbslone (dismissed):

frontend/packages/client/src/components/ProposalCreate/StepTwo.js
```diff
@@ -47,7 +47,10 @@ const StepTwo = ({ stepData, setStepValid, onDataChange }) => {
 
     // push now if date is today and not already in time interval
     if (cutOffDate) {
-      const nowDate = new Date();
+      const nowDate =
+        process.env.APP_ENV === 'PRODUCTION'
+          ? new Date(Date.now() + 60 * 60 * 1000) // delay by an hour in prod env
+          : new Date();
       nowDate.setSeconds(0);
```
germanurrus, line 54: I think this only applies for the ones that will make use of the snapshot process. 
If strategy for the proposal is `one-address-one-vote does` does it make sense to start the voting one hour later? what do you think @dbslone? This would require to see what strategy was selected on the previous step and checking the environment.

germanurrus, line 54: @fersan1985 I've checked on this file https://github.com/DapperCollectives/CAST/blob/ca706f83d8922e5a036e93816f24b3ff999cb8b1/frontend/Dockerfile.vars#L19-L20 and APP_ENV is not set in the docker build process. I do see NODE_ENV I think that's we should use instead.

germanurrus (changes requested):
Waiting for confirmation for APP env

frontend/packages/client/src/components/ProposalCreate/StepTwo.js
germanurrus: added APP_ENV on Dockerfile.stage file

germanurrus: App env added https://github.com/DapperCollectives/CAST/blob/c96756d62fee476ebba3e9a4d142329cee977626/frontend/Dockerfile#L12-L13

germanurrus:
@dbslone updates done, this is ready for new review now

sarahy-Dapper (approved):
REACT_APP_APP_ENV env variable is working as expected for development and production.

germanurrus (approved):

PR #139, opened 2022-07-12
cas-134 Display proposal strategy name instead of icon

CJamesKeller:
Changed icon to text
Screen Shot 2022-07-18 at 13 26 43

### DIFF
--- a/frontend/packages/client/src/components/ProposalInformation.js
+++ b/frontend/packages/client/src/components/ProposalInformation.js
@@ -121,7 +121,7 @@ const WrapperSpacingBottom = ({
 
 const ProposalInformation = ({
   creatorAddr = '',
-  strategy = {},
+  strategyName = '',
   isCoreCreator = false,
   ipfs = '',
   ipfsUrl = '',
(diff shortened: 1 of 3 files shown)


dbslone (approved):

frontend/packages/client/src/components/ProposalInformation.js
```diff
@@ -249,13 +249,13 @@ const ProposalInformation = ({
...
-            title={'Stategie(s)'}
+            title={strategies.length === 1 ? 'Strategy' : 'Strategies'}
             content={
               <div className="is-flex flex-1" onClick={openStrategyModal}>
                 {strategies.map((st, index) => {
                   return (
                     <div className="pr-2" key={`${st}-${index}`}>
-                      <StrategyIcon className="cursor-pointer" />
+                      {st.name}
                     </div>
```
germanurrus, line 259: Proposal will be using only one strategy for winner calculation, so maybe instead of using the map we could get the first element and show it instead of mapping 
```
const  [strategy] = strategies ?? [ { name: ''}];
```
Also use `Strategy` word instead of `Strategies` bc for now there will be only one.

Screen Shot 2022-07-13 at 17 37 54

germanurrus (changes requested):

PR #142, opened 2022-07-13
Rich/cas 219 setup CAST dev wallet with NFT tokens WIP

rvmelkonian:
- adds several new functions inside `/shared/flow.go` in order for the dev wallet to be loaded up with NFTs. This will enable the user to cast votes using the `balance-of-nfts` strategy.


backend/main/shared/flow.go
```diff
@@ -356,6 +421,30 @@ func WaitForSeal(
 	return result, tx, errTx
 }
 
+func (fa *FlowAdapter) CreateNFTCollection(ctx context.Context) {
```
jacksonConrad, line 424: where is this being called?
i just want to verify that it couldn't get run somehow in production, although it probably wouldn't work anyway.

jacksonConrad (approved):

backend/main/shared/flow.go
rvmelkonian: not being called anywhere yet, working on that part still

rvmelkonian:
closing this in favour of a simpler implementation on a new branch

PR #145, opened 2022-07-13
CAS-103 - fix join button position

germanurrus:
Applied style fixes for join community button, also avatar size was changed on mobile and some styling on community card based on CAS-103 feedback.

Before: 
Screen Shot 2022-07-14 at 12 14 34

Screen Shot 2022-07-14 at 12 14 46

After:
Screen Shot 2022-07-14 at 12 09 48
Screen Shot 2022-07-14 at 12 10 14

### DIFF
--- a/frontend/packages/client/src/components/Community/JoinCommunityButton.js
+++ b/frontend/packages/client/src/components/Community/JoinCommunityButton.js
@@ -34,7 +34,9 @@ export default function JoinCommunityButton({
     }
   }, [user?.addr, isModalErrorOpened, closeModal, setIsModalErrorOpened]);
 
-  const joinCommunity = async () => {
+  const joinCommunity = async (event) => {
+    event.preventDefault();
+    event.stopPropagation();
     if (!user?.addr) {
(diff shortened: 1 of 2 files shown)


frontend/packages/client/src/components/Community/JoinCommunityButton.js
```diff
@@ -62,7 +64,9 @@ export default function JoinCommunityButton({
     }
   };
 
-  const leaveCommunity = async () => {
+  const leaveCommunity = async (event) => {
+    event.preventDefault();
+    event.stopPropagation();
```
germanurrus, line 69: added this to prevent navigation from the community card to the community when clicking on join community in the home page

CJamesKeller: With the style changes, I agree that this is a better way than having it positioned `absolute`.

CJamesKeller (approved):

mannynotfound (approved):
Thank you!

PR #148, opened 2022-07-14
community page layout improvements

mannynotfound:
- tightens the space between "title" and "content" so association is more obvious. 
- remove "Details" header as it was always empty and redundant
- create shared `Title` component for community tabs
- adjust members spacing to match about spacing so content doesnt shift when navigating 

(grid ruler lines not part of changes, just shown for visual reference)
newabout
newmembers


frontend/packages/client/src/components/CommunityLinks.js
```diff
@@ -10,12 +11,7 @@ export default function CommunityLinks({
 } = {}) {
   return (
     <div className="columns my-0 is-multiline">
-      <div
-        className="column is-12 has-text-weight-bold is-uppercase small-text"
-        style={{ marginBottom: '20px' }}
-      >
-        Links
-      </div>
+      <Title>Links</Title>
```
germanurrus, line 14: this one need classes `column is-12` should Title allow classNames prop like `classNames={"column is-12"}`
Screen Shot 2022-07-14 at 13 58 47

mannynotfound:
@germanurrus updated to allow `className` on title 👍

frontend/packages/client/src/components/Title.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const Title = ({ children, className }) => (
```
germanurrus, line 3: I think this one needs to be initialized with `''` otherwise when nothing is passed it injects undefined in the class prop:
```
<h6 class="small-text has-text-weight-bold is-uppercase mb-1 undefined">About</h6>

Comment (germanurrus) [inline frontend/packages/client/src/components/Title.js:3]:
```suggestion
const Title = ({ children, className = '' } = {}) => (
```

germanurrus: was this pushed? I don't see it in the branch

mannynotfound:
@germanurrus should be g2g now

PR #154, opened 2022-07-14
CAS-229: Add current user to Leaderboard

jbluks:
**Ticket:** [CAS-229](https://linear.app/dappercollectives/issue/CAS-229/add-current-user-to-leaderboard)

**Description:**

Adds Current User to the Leaderboard, if wallet is connected. Leaderboard API now returns a user if passed an address.

Screen Shot 2022-07-14 at 4 26 26 PM


dbslone:
When I try to cast a vote on a community that I just created I receive the error `ERROR: relation "user_achievements" does not exist (SQLSTATE 42P01)`

jbluks:
Checking it out - it seems like the database/migration are out of sync and that has created an issue. I have dropped everything locally, ran the migrations and voted via the UI and it all worked as expected.

frontend/packages/client/src/hooks/useLeaderBoard.js
```diff
@@ -2,28 +2,27 @@ import { useQuery } from 'react-query';
...
+    ['leaderboard', addr],
+    async () => fetchLeaderBoard(communityId, addr)
   );
 
   if (isError) {
     notifyError(error);
   }
 
+  const leaderBoard = data?.data?.Users ?? [];
+  const currentUser = data?.data?.CurrentUser;
```
germanurrus, line 17: can we make this `users` and `currentUsers` ?

jbluks: Done - for currentUser, there will only ever be one user since that is the user whose wallet is connected to CAST. There will be no currentUser if wallet is not connected.

germanurrus: my bad, it would be `currentUser`

jbluks: Updated backend with json tag to return `users`/`currentUser`

frontend/packages/client/src/hooks/useLeaderBoard.js
```diff
@@ -2,28 +2,27 @@ import { useQuery } from 'react-query';
...
+    ['leaderboard', addr],
+    async () => fetchLeaderBoard(communityId, addr)
   );
 
   if (isError) {
     notifyError(error);
   }
 
+  const users = data?.data?.users ?? [];
+  const currentUser = data?.data?.current_user;
```
germanurrus, line 17: could this be like 
`const currentUser = data?.data?.currentUser;` 
so it's camelcase?

jbluks: Upated.

germanurrus (approved):
LGTM! (frontend changes) thanks for the updates @jbluks

dbslone (approved):

PR #155, opened 2022-07-14
fix contract validation when form is empty

germanurrus:


frontend/packages/client/src/components/CommunityCreate/StepThree.js
```diff
@@ -10,9 +10,8 @@ const allEmptyFields = (data) => {
     'storagePath',
     'proposalThreshold',
   ];
-  return (
-    fields.every((field) => data[field] === undefined) ||
-    fields.every((field) => data[field] === '')
+  return fields.every(
+    (field) => data[field] === undefined || data[field] === ''
   );
```
germanurrus, line 15: if form was touched in some fields, then some fields will have '' and the ones untouched will be  undefined. Covering cases where all are untouched (undefined), all touched but empty strings ('') and a mix of the two conditions named before.

dbslone (approved):

PR #158, opened 2022-07-15
make create community steps the same color scheme as create proposal

mannynotfound:
the colors for the steps while creating a community should be the same as when creating a proposal. this PR removes the customization option of the step colors in StepByStep to make them consistent across all uses


germanurrus (approved):
LGTM!

PR #171, opened 2022-07-15
CAS-207: Update GetUserCommunities to return single community with multiple roles

jbluks:
Ticket: [CAS-207](https://linear.app/dappercollectives/issue/CAS-207/update-backend-to-not-create-duplicate-user-records)

**Description:**

Backend call was returning duplicate communities for user with multiple roles in the community. Updated query to combine roles into array for single community.

e.g.

```
{
    "data": [
        {
            "id": 1,
            "name": "Flow",
            "body": "Vote on Flow Validators",
            "strategies": [
                {
                    "name": "one-address-one-vote",
                    "contract": {
                        "name": "FlowToken",
                        "addr": "0x0ae53cb6e3f42a79",
                        "publicPath": "flowTokenBalance",
                        "threshold": "0"
                    }
                }
            ],
            "twitterUrl": "https://twitter.com/flow_blockchain",
            "discordUrl": "https://discord.gg/flow",
            "onlyAuthorsToSubmit": true,
            "timestamp": "",
            "compositeSignatures": null,
            "creatorAddr": "0xf8d6e0586b0a20c7",
            "createdAt": "2022-07-15T17:04:45.707396Z",
            "Roles": "admin,author,member"
        }
    ],
    "start": 0,
    "count": 1,
    "totalRecords": 1,
    "next": -1
}
```


backend/main/models/community_users.go
```diff
@@ -30,7 +31,12 @@ var USER_TYPES = UserTypes{"member", "author", "admin"}
 
 type UserCommunity struct {
 	Community
-	Membership_type string `json:"membershipType,omitempty"`
+	Roles string
+}
+
+type UserCommunityPayload struct {
+	Community
+	Roles []string `json:"roles" validate:"required"`
```
germanurrus, line 39: Can we make it return `roles:` instead of `Roles` to keep it the same with the other fields the frontend is getting as response?

germanurrus (changes requested):

germanurrus:
@jbluks could you also update this function https://github.com/DapperCollectives/CAST/blob/main/frontend/packages/client/src/hooks/useUserRoleOnCommunity.js#L46

to use this code:
```
  const rolesInCommunity =
    communityUser
      ?.find((datum) => {
        return datum.id.toString() === communityId.toString();
      })
      ?.roles?.split(',') ?? [];
```
so the FE can split the roles string, this would be the only thing required to make this backend change to work.

germanurrus (approved):
Frontend code looks good!

rvmelkonian:
looks good !

PR #172, opened 2022-07-15
Updated readme

tuffon:


backend/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Prerequisites
 
-- [GoLang 1.6](https://golang.org/doc/install)
+- [GoLang 1.16](https://golang.org/doc/install)
```
dbslone, line 5: i don't think this is needed I am running 1.18 and we have other ppl running 1.16.

backend/README.md
```diff
@@ -87,10 +98,14 @@ make testmigratedown
 
 #### Dealing with Dirty Schema Migrations
 
+Reset Schema Migrations
+```bash
+UPDATE schema_migrations SET dirty = false
+```
```
dbslone, line 104: thanks for adding this to the doc

backend/README.md
```diff
@@ -87,10 +98,14 @@ make testmigratedown
...
 
 ### Testing
 
-1. Run `flow emulator` from this top-level directory (the private key for the `emulator-account` in `flow.json` must match the private key hard-coded in the test suite). Install the `flow` CLI [here](https://docs.onflow.org/flow-cli/install/)
+1. Run `flow emulator --dev-wallet` from this top-level directory (the private key for the `emulator-account` in top level `flow.json` must match the private key hard-coded in the test suite eg. `backend/cadence/V2/tests/flow.json`). Install the `flow` CLI [here](https://docs.onflow.org/flow-cli/install/)
```
dbslone, line 108: I don't think this is required for backend tests. @rvmelkonian @jacksonConrad can you comment on this

frontend/README.md
```diff
@@ -4,7 +4,7 @@
 
 - [Yarn](https://classic.yarnpkg.com/lang/en/docs/install)
 - [GoLang 1.6](https://golang.org/doc/install)
-- [Node/NPM](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
+- [Node/NPM v16](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
```
dbslone, line 7: 👍 Thanks for adding the node version. I'm wondering if we should include `nvm` instructions so its easy to upgrade/downgrade node versions

tuffon: sure i'll update PR with the nvm instructions

backend/README.md
rvmelkonian: yes that's correct actually its only required as a flag when running the front end, so maybe best to leave it as an optional flag for now

rvmelkonian:
looks like some great improvements here!

PR #173, opened 2022-07-15
CAS-244: Upgrade to FCL v1.1.0

jacksonConrad:
* Upgrade to FCL v1.1.0
* Update FCL Config to include Ledger wallet support
* Updates `yarn chain` command to work with Flow CLI v0.37.5, which now supports `dev-wallet` command
* Updates README to account for new Flow CLI version
* Leaves Dapper Wallet address in `networks.js` commented out until we are ready to include it.


germanurrus (commented):
Tested locally signing requests and works solid!

frontend/README.md
```diff
@@ -18,7 +17,7 @@ yarn install
 
 #### Contracts
 
-To start a local blockchain using `flow emulator`:
+To start a local blockchain & dev-wallet:
```
mannynotfound, line 20: it looks like the `yarn chain` command doesnt work  anymore and that script is now in `/backend` so these instructions should be in the backend repo?

rvmelkonian:
is the plan to have this merged in before launch? Tbh I'm a bit nervous about doing this at the same time we launch. Having some kind of integration / cypress tests to test out the UI would ease the fear.

jacksonConrad:
> is the plan to have this merged in before launch? Tbh I'm a bit nervous about doing this at the same time we launch. Having some kind of integration / cypress tests to test out the UI would ease the fear.

yes i think we want it out ASAP because otherwise we have Dapper Wallet showing up as an option which we don't want yet.

dbslone (approved):

PR #174, opened 2022-07-15
fix loading admins

germanurrus:


frontend/packages/client/src/pages/Community.js
```diff
@@ -417,7 +417,7 @@ export default function Community() {
                 {activeTabMap['proposals'] && (
                   <CommunityProposals
                     communityId={community.id}
-                    admins={admins}
+                    admins={admins ?? []}
```
germanurrus, line 420: while loading admins is `null` and makes UI to break. Will be refactored when using react-query on the hook that calls the api

dbslone (approved):

PR #177, opened 2022-07-18
github action move path check to PR only

dbslone:


dbslone:
if pr check is required and path for backend doesn't change then pr checks done pass need to examine more how to resolve this problem

PR #178, opened 2022-07-18
CAS-248 - add error message for in-line image

germanurrus:
Added Error Message on upload:
Screen Shot 2022-07-18 at 11 58 51

Not Related but fixes style on mobile:
Before:
Screen Shot 2022-07-18 at 12 00 20

After:
Screen Shot 2022-07-18 at 12 16 29

### DIFF
--- a/frontend/packages/client/src/components/UploadImageModal.js
+++ b/frontend/packages/client/src/components/UploadImageModal.js
@@ -1,8 +1,11 @@
 import React, { useCallback, useState, useEffect, useMemo } from 'react';
 import { useDropzone } from 'react-dropzone';
-import { useFileUploader } from '../hooks';
-import { Upload, Bin } from '../components/Svg';
-import { Loader } from '../components';
+import { useFileUploader, useMediaQuery } from 'hooks';
+import { Upload, Bin } from 'components/Svg';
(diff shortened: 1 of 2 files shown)


frontend/packages/client/src/components/UploadImageModal.js
```diff
@@ -134,12 +139,13 @@ export default function UploadImageModal({
   onDismiss,
   isCancelling = false,
   onDone,
+  maxImageFiles = MAX_IMAGE_FILES,
```
germanurrus, line 142: made this a prop for the component

frontend/packages/client/src/const/index.js
```diff
@@ -3,6 +3,8 @@ export const MAX_FILE_SIZE = 5242880;
 // 1024 * 1024 * 2 = 2MB
 export const MAX_AVATAR_FILE_SIZE = 2097152;
 
+export const MAX_PROPOSAL_IMAGE_FILE_SIZE = 2097152;
+
```
germanurrus, line 7: this is kind of repeated but applies only for in-line images updated to proposals

mannynotfound (approved):
LGTM!

PR #179, opened 2022-07-18
animate active bar under community tab links

mannynotfound:
Animates the "active" underline underneath a navigation tab on the community page similar to https://codepen.io/everdimension/pen/xZLggo but with support for dynamic tab link widths / positions.

### DIFF
--- a/frontend/packages/client/src/App.sass
+++ b/frontend/packages/client/src/App.sass
@@ -300,7 +300,7 @@ hr
 	border-bottom-width: 3px
 	border-bottom-style: solid
 	&.tab-community
-		border-bottom-color: $light-green
+		border-bottom-color: transparent
 		color: $black
 		border-bottom-width: 2px
 		border-bottom-style: solid
--- a/frontend/packages/client/src/pages/Community.js
+++ b/frontend/packages/client/src/pages/Community.js
@@ -19,6 +19,7 @@ import {
   useCommunityUsers,
   useUserRoleOnCommunity,
   useCommunityMembers,
+  useWindowDimensions,
 } from '../hooks';
 import { useWebContext } from '../contexts/Web3';
 import Blockies from 'react-blockies';
@@ -137,6 +138,7 @@ export default function Community() {
 
   const history = useHistory();
 
+  const { width: windowWidth } = useWindowDimensions();
   const [activeWidth, setActiveWidth] = useState(105);
   const [activeLeft, setActiveLeft] = useState(0);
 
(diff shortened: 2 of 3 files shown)


frontend/packages/client/src/App.sass
```diff
@@ -300,7 +300,7 @@ hr
 	border-bottom-width: 3px
 	border-bottom-style: solid
 	&.tab-community
-		border-bottom-color: $light-green
+		border-bottom-color: transparent
```
mannynotfound, line 303: make this transparent as we will use our own custom underline element

frontend/packages/client/src/pages/Community.js
```diff
@@ -140,6 +140,9 @@ export default function Community() {
 
   const history = useHistory();
 
+  const [activeWidth, setActiveWidth] = useState(105);
```
mannynotfound, line 143: 105 is a little magic number here but its the default width of the default tab (proposals). We could make this dynamic but for sake of simplicity went with this

frontend/packages/client/src/pages/Community.js
```diff
@@ -335,13 +360,29 @@ export default function Community() {
...
+                      style={{
+                        transition: '0.1s all',
+                        position: 'absolute',
+                        bottom: 0,
+                        height: 2,
+                        backgroundColor: '#00EF8B',
+                        width: activeWidth,
+                        left: activeLeft,
+                      }}
+                    />
```
mannynotfound, line 378: this is our underline element that transitions its width and left offset smoothly

frontend/packages/client/src/pages/Community.js
```diff
@@ -196,6 +199,27 @@ export default function Community() {
...
+    };
+    const el = refMap[activeTab].current;
+    if (!el) return;
+
+    const elRect = el.getBoundingClientRect();
+    const parentRect = wrapperRef.current.getBoundingClientRect();
+    const offsetLeft = elRect.left - parentRect.left;
+    setActiveWidth(el.offsetWidth);
+    setActiveLeft(offsetLeft);
+  }, [activeTab]);
```
germanurrus, line 221: This looks really cool!! maybe you can add `windowWidth` from `const { width: windowWidth } = useWindowDimensions(); ` so when resizing the activeLeft and  activeLeft values are updated dynamically to prevent wrong positioning unit component is refreshed

frontend/packages/client/src/pages/Community.js
rvmelkonian: is this ok for mobile devices etc? If it's filling the width of the tab I'm guessing yes ..

PR #185, opened 2022-07-19
Get Contract by matching strategy name

dbslone:


rvmelkonian:
I refactored this a bit so that now `FetchBalance` only takes in two params, 
```
FetchBalance(b *models.Balance, p *models.Proposal) (*models.Balance, error)
```
It gets access to the DB and snapshot client using the separate method `InitStrategy` which you can think of like a constructor, calling `InitStrategy`, allows the DB and Snapshot Client to be accesses within the Strategy methods, without having to pass in those parameters, so a bit cleaner.

rvmelkonian (approved):

PR #188, opened 2022-07-19
update docker config

germanurrus:
- Upgrade alpine version 16.3.0
- Using yarn to build front end and using workspace build command
- Copying yarn.lock file


fersan1985:
As I told you Ger, we have to take staging as a base and add these changes, then GHA will put all this to Staging and Production. @germanurrus

fersan1985 (approved):

PR #192, opened 2022-07-19
CAS-223 Add character limit/character count to proposal title field

germanurrus:
Screen Shot 2022-07-19 at 17 58 52


frontend/packages/client/src/components/Popover.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import { useMediaQuery } from 'hooks';
 import classnames from 'classnames';
-import { map } from 'lodash';
```
germanurrus, line 4: unrelated change

dbslone (approved):

PR #193, opened 2022-07-20
cas-219 setup dev wallet with NFT tokens

rvmelkonian:
This PR allows the dev wallet to own an NFT of `ExampleNFT.cdc`, this should enable smoke testing `balance-of-nfts` strategy in local dev. 
* upgrades to Go 1.18
* upgrades overflow and packages
* merges in updates from cas-272. Did this because of breaking changes in overflow upgrade. 
* adds new ExampleNFT (post-secure-cadence)
* adds new ExampleNFT scripts 
* creates new method `processTokenThreshold` to handle both FT and NFT types for balance lookup. Before threshold was only handling the FT case.
* updates `EnforceTokenThreshold` for the same reason above. 
* updates `GetNFTIds` to parse script return value as type `[]interface{}` this means we can handle both string-slice(uuid) and uint64-slice ID types
* updates makeFile
* updates flow.json

Run `make nfts` to deploy nft contracts and execute the necessary transactions in the emulator. The NFT code is from the latest updates on the flow repo and requires the latest version of the flow-cli to compile. To upgrade flow-cli run `brew upgrade flow-cli`

It takes some time for these transactions to process and become `SEALED` on the emulator, this only occurs on the first run. After that caching kicks in.


backend/Makefile
```diff
@@ -14,16 +14,28 @@ test:
...
+	flow accounts add-contract FLOATVerifiers ./main/cadence/float/FLOATVerifiers.cdc
 
 macdev:
 	APP_ENV=DEV make macrun
 
 dev:
 	APP_ENV=DEV make build_and_run
 
-migrateup:
-	migrate -path ./migrations -database "postgres://postgres:admin@localhost:5432/flow_snapshot?sslmode=disable" -verbose up
```
jacksonConrad, line 26: did you mean to delete `migrateup`? :)

rvmelkonian: haha nop! will add in now

backend/main/server/app.go
```diff
@@ -589,10 +589,10 @@ func (a *App) createVoteForProposal(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	// validate user signature
-	if err := a.FlowAdapter.UserTransactionValidate(v.Addr, v.Message, v.Composite_signatures, v.TransactionId, a.TxOptionsAddresses, p.Choices); err != nil {
```
jacksonConrad, line 592: we will want to resolve the issue here before merging because otherwise Ledger wallets won't be validating properly

jacksonConrad (changes requested):
just need to get the UserTransactionValidate function working again.  The package upgrade broke this because the `jsoncdc.Validate` function is expecting an additional arg.

rvmelkonian:
sweet. working on it.

jacksonConrad (approved):

PR #200, opened 2022-07-20
update variables

germanurrus:


frontend/Dockerfile
```diff
@@ -9,10 +9,10 @@ COPY ./packages/client/ ./
 WORKDIR /app
 RUN yarn install
 
+#ENVS FOR PROD FRONTEND
 ENV APP_ENV "production"
 ENV NODE_ENV "production"
 ENV REACT_APP_FRONTEND_URL "cast.fyi"
-ENV NODE_ENV "production"
```
germanurrus, line 15: duplicated from line 12

fersan1985: On it!

PR #203, opened 2022-07-21
CAS-249: Add ability to return specific communities for homepage

jbluks:
**Ticket:** [CAS-249](https://linear.app/dappercollectives/issue/CAS-249/add-ability-to-return-specific-communities-for-homepage)

**Description:**

- Adds boolean field to Communities table `is_featured` which can be set to indicate the community is a featured community.
- `/communities-for-homepage` query checks for the `is_featured` flag


backend/main/models/community.go
```diff
@@ -156,8 +157,8 @@ func GetCommunitiesForHomePage(db *s.Database, start, count int) ([]*Community,
     	WHERE status = 'published' AND end_time < (NOW() AT TIME ZONE 'UTC')
     	GROUP BY community_id
     	HAVING COUNT(*) >= 2
-  	)
-	OR id = 1
+  	))
+	OR is_featured = 'true'
```
jbluks, line 161: @dbslone For this query, did we only want to return featured communities, OR keep the existing logic as well, OR, only return communities that match the existing logic AND are featured.

dbslone: This is good I can update the Flow community to have the featured flag so it appears

dbslone (approved):

PR #207, opened 2022-07-21
Update error messages to sentence case and punctuation

tuffon:


rvmelkonian:
we like nice error messages. I think you just need to pull from main and resolve those conflicts then should be good!

dbslone (approved):

PR #213, opened 2022-07-21
conditional check

dbslone:


frontend/packages/client/src/App.js
```diff
@@ -17,7 +17,6 @@ const isLocalDevMode = process.env.NODE_ENV === 'development';
 const queryClient = new QueryClient();
 
 function App() {
-  console.log({ FLOW: process.env.REACT_APP_FLOW_ENV });
```
germanurrus, line 20: should we console.log this?

PR #219, opened 2022-07-21
CAS-314 update border on create community button

germanurrus:


frontend/packages/client/src/components/ActionButton.js
```diff
@@ -10,7 +10,7 @@ export default function ActionButton({
   classNames,
 } = {}) {
   const clNames = classnames(
-    'button transition-all is-flex is-align-items-centered rounded-sm is-uppercase',
```
germanurrus, line 13: removing `transition-all` to remove the animation on the border appearing

dbslone (approved):

PR #220, opened 2022-07-22
fix environment variable

germanurrus:
- Fix environment variable


frontend/packages/client/src/pages/CommunityCreate.js
```diff
@@ -99,8 +99,9 @@ export default function CommunityCreate() {
           // This is to bypass error on local
           // emulator when keys field is not present
           // on flow emulator response
-          if (
-            process.env.APP_ENV !== 'PRODUCTION' &&
```
germanurrus, line 103: without the REACT_APP prefix this is not working

dbslone (approved):

PR #229, opened 2022-07-22
cas-298 - add component and ref to calculate width

germanurrus:
Shrink author address to fit in container:

https://user-images.githubusercontent.com/7526740/180472627-c0c70a68-a512-4865-8b25-94445d78f0ab.mov

### DIFF
--- a/frontend/packages/client/src/components/ProposalInformation.js
+++ b/frontend/packages/client/src/components/ProposalInformation.js
@@ -6,20 +6,72 @@ import { useVotingResults, useWindowDimensions } from '../hooks';
 import useMediaQuery, { mediaMatchers } from '../hooks/useMediaQuery';
 import Tooltip from './Tooltip';
 
-const InfoBlock = ({ title, content }) => {
+function truncate(str) {
+  return str.substr(0, 3) + '...' + str.substr(10, str.length);
+}
+
+const BlockieWithAddress = React.forwardRef(
+  ({ creatorAddr, isCoreCreator }, ref) => {
+    const [addr, setAdd] = useState(creatorAddr);
+
+    const { width } = useWindowDimensions();
+
+    useEffect(() => {
+      if (ref?.current.clientWidth <= 260 && creatorAddr === addr) {
+        setAdd(truncate(creatorAddr));
+      } else if (ref?.current.clientWidth > 260 && creatorAddr !== addr) {
+        setAdd(creatorAddr);
+      }
+    }, [ref, width, creatorAddr, addr]);
+
+    return (
+      <div className="columns is-mobile m-0">
+        <div className="column is-narrow is-flex is-align-items-center p-0">
+          <Blockies
+            seed={creatorAddr}
+            size={6}
+            scale={4}
+            className="blockies"
+          />
+        </div>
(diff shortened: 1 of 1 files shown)


frontend/packages/client/src/components/ProposalInformation.js
```diff
@@ -6,20 +6,72 @@ import { useVotingResults, useWindowDimensions } from '../hooks';
...
+  }
   return (
-    <div className="columns is-mobile p-0 m-0 mb-5 small-text">
-      <div
-        className="column p-0 is-flex is-align-items-center flex-1 has-text-grey"
-        style={{ maxWidth: '40%' }}
-      >
+    <div
+      className="columns is-mobile p-0 m-0 mb-5 small-text"
+      ref={containerRef}
```
germanurrus, line 62: creating ref so children knows parent size to know if it needs to shrink

frontend/packages/client/src/components/ProposalInformation.js
```diff
@@ -258,28 +310,11 @@ const ProposalInformation = ({
...
-                <div className="column is-narrow is-flex is-align-items-center p-0">
-                  <Blockies
-                    seed={creatorAddr}
-                    size={6}
-                    scale={4}
-                    className="blockies"
-                  />
-                </div>
-                <div className="column px-2 py-0 is-flex flex-1 is-align-items-center">
-                  {creatorAddr}
```
germanurrus, line 272: can't use something like:
```
max-width: 150px;
overflow: hidden;
white-space: nowrap;
text-overflow: ellipsis;
display: inline-block;
```
bc size depends on parent component and setting a max with at this level doesn't work, component needs to know parent size to shrink or not

frontend/packages/client/src/components/ProposalInformation.js
```diff
@@ -6,20 +6,72 @@ import { useVotingResults, useWindowDimensions } from '../hooks';
...
+}
+
+const BlockieWithAddress = React.forwardRef(
+  ({ creatorAddr, isCoreCreator }, ref) => {
+    const [addr, setAdd] = useState(creatorAddr);
+
+    const { width } = useWindowDimensions();
+
+    useEffect(() => {
+      if (ref?.current.clientWidth <= 223 && creatorAddr === addr) {
```
mannynotfound, line 20: this number might vary depending on the contents of the address, for example if there is a lot of W (the widest letter) vs a lot of I and 1 etc - probably works well enough for most addresses though or maybe add a bit of extra padding (250?) in case of longer addresses. Otherwise youll want to give the address `white-space: nowrap` and then calculate if the width of the address is greater than the width of the parent container and then truncate

frontend/packages/client/src/components/ProposalInformation.js
mannynotfound: another strategy do this with text-overflow would be to set `maxWidth` dynamically with javascript using refs etc but I think truncating an address by removing the middle chunks is much better than removing the end (people often use the first/last four characters of an address as a quick identifier)

mannynotfound (approved):
I think this is good enough for now given how minor of an issue it is but width could be calculated more dynamically

PR #248, opened 2022-07-22
CAS-315: Fix Leaderboard showing current user if on leaderboard

jbluks:
**Ticket:** [CAS-315](https://linear.app/dappercollectives/issue/CAS-315/leaderboard-populates-duplicate-addresses)

**Description:**

- Current User was being displayed when present in Leaderboard. Updated logic to check if user exists on leaderboard and only displays user if they are not already on the leaderboard.


mannynotfound:
this work is included in https://github.com/DapperCollectives/CAST/pull/239

PR #262, opened 2022-07-25
cas 240 - Updated Community Page Layout

germanurrus:
based on https://github.com/DapperCollectives/CAST/pull/261

Update style:

Screen Shot 2022-07-25 at 11 51 09
Screen Shot 2022-07-25 at 11 50 15

### DIFF
--- a/frontend/packages/client/src/App.sass
+++ b/frontend/packages/client/src/App.sass
@@ -136,6 +136,8 @@ code
 	border-bottom-right-radius: 8px
 .rounded-lg
 	border-radius: 20px
+.rounded-extra-lg
+	border-radius: 40px
 .rounded-lg-top
 	border-top-left-radius: 20px
 	border-top-right-radius: 20px
@@ -352,7 +354,7 @@ hr
 	padding: 19.5px 24px
 	max-width: 225px
 .community-content
-	padding-top: 31px
+	padding-top: 50px
 .proposal-title
 	font-size: 18px
 	line-height: 120%
@@ -420,9 +422,8 @@ hr
 		padding: 24px 24px 0px 24px
 	.community-content
 		padding-top: 0
-		margin-top: -70px
 	.community-header-wrapper
-		max-height: 288px
+		max-height: 152px
 	.community-header
 		flex-direction: column
 		min-height: 288px
--- a/frontend/packages/client/src/App.sass
+++ b/frontend/packages/client/src/App.sass
@@ -136,7 +136,7 @@ code
 	border-bottom-right-radius: 8px
 .rounded-lg
 	border-radius: 20px
-.rounded-extra-lg
+.rounded-xl
 	border-radius: 40px
 .rounded-lg-top
 	border-top-left-radius: 20px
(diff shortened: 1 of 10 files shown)


frontend/packages/client/src/App.sass
```diff
@@ -136,6 +136,8 @@ code
 	border-bottom-right-radius: 8px
 .rounded-lg
 	border-radius: 20px
+.rounded-extra-lg
```
mannynotfound, line 139: nitpick but this can be `rounded-xl`

frontend/packages/client/src/components/Community/JoinCommunityButton.js
```diff
@@ -84,19 +88,25 @@ export default function JoinCommunityButton({
   };
 
   const classNamesContainer = classnames(
-    'column is-narrow-tablet is-full-mobile is-align-self-center',
+    'column is-narrow-tablet is-full-mobile',
     {
       [classNames]: !!classNames,
+      [containerAlignment]: !!containerAlignment,
```
mannynotfound, line 94: since these are initialized by default they should never be undefined so you can just do 

```
const classNamesContainer = `column is-narrow-tablet is-full-mobile `${classNames} ${containerAligntment}`
```

frontend/packages/client/src/components/Community/JoinCommunityButton.js
```diff
@@ -84,19 +88,25 @@ export default function JoinCommunityButton({
...
     }
   );
 
   const classNamesButton = classnames('button is-uppercase is-fullwidth', {
-    'rounded-lg has-background-black has-text-white-bis': darkMode,
+    'has-background-black has-text-white-bis': darkMode,
     'rounded-sm has-text-black border-lighter-dark-grey small-text': !darkMode,
+    [buttonClassNames]: !!buttonClassNames,
+    [borderRadious]: !!borderRadious,
```
mannynotfound, line 102: same comment as above these can just be added as variables

frontend/packages/client/src/pages/Community.js
```diff
@@ -256,111 +255,33 @@ export default function Community() {
...
+            setTotalMembers={setTotalMembers}
+            onLeaveCommunity={onUserLeaveCommunity}
+            onJoinCommunity={onUserJoinCommunity}
+            classNames="small-text pb-0-mobile px-0-mobile"
+            extraStyles={notMobile ? { maxWidth: '132px' } : {}}
+            buttonClassNames="py-5 py-4-mobile"
+            borderRadious="rounded-extra-lg"
+            containerAlignment="is-align-items-flex-start"
+          />
+        }
```
mannynotfound, line 283: is `joinCommunityButton` expected to be dynamic? otherwise seems like this component can just live in `CommunityHeader`

mannynotfound (approved):
LGTM, few comments but nothing blocking

frontend/packages/client/src/pages/Community.js
germanurrus: I did this to avoid passing props for the JoinButton which is using these props: 
```
 communityId={communityId}
 setTotalMembers={setTotalMembers}
 onLeaveCommunity={onUserLeaveCommunity}
 onJoinCommunity={onUserJoinCommunity}
 ```
 but totally fine with passing down this props, what do you think?

germanurrus: Ended up moving the component inside the CommunityHeader component

PR #275, opened 2022-07-25
fix strategy update on community creation

germanurrus:
- Fix loader position on edit community
- Allow to delete strategy on community creation when there is only one strategy. In case the information entered for the strategy is not valid, user needs to be able to delete the existing one to be able to add it again to fix the strategy data entered.

### DIFF
--- a/frontend/packages/client/src/components/CommunityCreate/StepFour.js
+++ b/frontend/packages/client/src/components/CommunityCreate/StepFour.js
@@ -27,7 +27,7 @@ export default function StepFour({
     <StrategySelectorForm
       existingStrategies={strategies}
       onStrategySelection={onStrategySelection}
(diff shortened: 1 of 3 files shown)


frontend/packages/client/src/components/CommunityCreate/StepFour.js
```diff
@@ -27,6 +27,7 @@ export default function StepFour({
     <StrategySelectorForm
       existingStrategies={strategies}
       onStrategySelection={onStrategySelection}
+      enableDelUniqueItem={true}
```
mannynotfound, line 30: this can just be `enableDelUniqueItem` which will make it true by default

mannynotfound (approved):

PR #285, opened 2022-07-25
CAS-345: Fix UserCommunities query to return proper roles

jbluks:
**Ticket:** [CAS-345](https://linear.app/dappercollectives/issue/CAS-345/when-viewing-a-proposal-page-im-seeing-the-cancel-proposal-button-when)

**Description:**

Query for returning merging roles in Community was incorrect. Fixed to Query the DB for all user communities, then merge multiple communities into 1 with list of unique roles.

Note: Also fixes some broken test based on string changes

### DIFF
--- a/backend/main/models/community_users.go
+++ b/backend/main/models/community_users.go
@@ -145,25 +145,12 @@ func GetCommunitiesForUser(db *s.Database, addr string, start, count int) ([]Use
 	err := pgxscan.Select(db.Context, db.Conn, &communities,
 		`
 		SELECT
-	  communities.*,
-	  a.roles
-		FROM communities
-		LEFT JOIN community_users ON community_users.community_id = communities.id
-		LEFT JOIN (
-				SELECT DISTINCT ON (community_users.community_id) community_users.*,
-						ARRAY(
-								SELECT community_users.user_type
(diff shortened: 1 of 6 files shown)


backend/main/models/community_users.go
```diff
@@ -145,59 +145,22 @@ func GetCommunitiesForUser(db *s.Database, addr string, start, count int) ([]Use
...
-								WHERE addr = $1
-								GROUP BY community_users.user_type
-						) as roles
-						FROM community_users
-						WHERE addr = $1
-		) a ON community_users.addr = a.addr
+		communities.*,
+		community_users.user_type as roles
+		  FROM communities
+		  LEFT JOIN community_users ON community_users.community_id = communities.id
```
jacksonConrad, line 151: i believe this should be a regular `JOIN` rather than a `LEFT JOIN`.  This will select every community from the DB, and as this number grows, the result could get very large

jacksonConrad (changes requested):

backend/main/models/community_users.go
jbluks: Fixed.

dbslone (approved):

PR #290, opened 2022-07-26
CAS-355 Bugfix: Upgrade to @onflow/fcl@alpha to fix /v1/accounts calls

jacksonConrad:
**Explanation**

There's an issue @Sarah Younes flagged when creating communities where the FE is validating the provided addresses for the community strategies, and the FCL call is intermittently returning 400 errors and blocking the community creation.

It looks like the reason is because on our version of FCL, the /v1/accounts call is specifying block_height=final .  This is a problem because if the Execution Node isn't fully synced, then the account wont "exist" at the final block-height.

They made a fix for this 12 days ago in the fcl-js repo (the solution is to specify block_height=sealed which is much more reliable) but we need to upgrade to the FCL alpha release by doing npm install @onflow/fcl@alpha

**Solution**
`npm install @onflow/fcl@alpha`


frontend/packages/client/package.json
```diff
@@ -32,8 +32,8 @@
     "react-query": "^3.39.1",
     "react-router-dom": "^5.2.0",
     "react-scripts": "4.0.3",
-    "string-strip-html": "8.3.0",
-    "sass": "^1.49.8"
+    "sass": "^1.49.8",
+    "string-strip-html": "8.3.0"
   },
   "scripts": {
     "start": "react-scripts start",
```
germanurrus, line 39: can you remove `package-lock.json` and run `yarn install` ?

germanurrus (changes requested):

frontend/packages/client/package.json
jacksonConrad: sure thing

germanurrus (approved):
Tested locally works as expected

dbslone:
Closing PR instead will use #317

PR #291, opened 2022-07-26
add star animation

germanurrus:
Added animation to home page:

https://user-images.githubusercontent.com/7526740/181067141-24316e71-b598-4cac-8d91-35e6b3f13db5.mov


frontend/packages/client/src/components/HomeHeader.js
```diff
@@ -1,17 +1,44 @@
...
+  { topPer: 0.1, leftPer: 0.45, width: '31px', height: '31px', start: 700 },
+  { topPer: 0.1, leftPer: 0.99, width: '35px', height: '35px', start: 1100 },
+  { topPer: 0.35, leftPer: 0.7, width: '31px', height: '31px', start: 1100 },
+  { topPer: 0.9, leftPer: 0.7, width: '25px', height: '25px', start: 500 },
+];
 
 export default function HomeHeader() {
+  const { addToArrayOfRefes, starArray, parentRef } = useStarAnimation({
+    stars,
+  });
```
mannynotfound, line 17: probably makes more sense to initialize `parentRef` in `HomeHeader` and then pass it to `useStarAnimation` as an argument. then `useStarAnimations` would just return `starArray` which you can render in the JSX wherever you want

frontend/packages/client/src/hooks/useStarAnimation.js
```diff
@@ -0,0 +1,55 @@
...
+      setStarArray(
+        stars.map((position) => ({
+          width: position.width,
+          height: position.height,
+          top: maxHeight * position.topPer,
+          left: maxWidth * position.leftPer,
+        }))
+      );
+    }
+  }, [parent, width, height, stars]);
```
germanurrus, line 33: added width, height so when resizing stars can change with screen size. Another option would be to provide three arrays ( desktop, tablet, mobile)

frontend/packages/client/src/components/HomeHeader.js
```diff
@@ -1,17 +1,44 @@
...
+  { topPer: 0, leftPer: 0, width: '31px', height: '31px', start: 100 },
+  { topPer: 0.42, leftPer: 0.18, width: '20px', height: '20px', start: 1000 },
+  { topPer: 0.1, leftPer: 0.45, width: '31px', height: '31px', start: 700 },
+  { topPer: 0.1, leftPer: 0.99, width: '35px', height: '35px', start: 1100 },
+  { topPer: 0.35, leftPer: 0.7, width: '31px', height: '31px', start: 1100 },
+  { topPer: 0.9, leftPer: 0.7, width: '25px', height: '25px', start: 500 },
+];
 
 export default function HomeHeader() {
+  const { addToArrayOfRefes, starArray, parentRef } = useStarAnimation({
```
mannynotfound, line 15: should probably be `addToArrayOfRefs`

mannynotfound (approved):
LGTM! left a thought on code structure but not a blocker IMO

PR #311, opened 2022-07-27
cas 350 - Update proposal routes

germanurrus:
- Update route configuration for proposal url


frontend/packages/client/src/pages/index.js
```diff
@@ -41,9 +44,6 @@ export default function AppPages() {
             <Route exact path="/proposal/create">
               <ProposalCreate />
             </Route>
-            <Route path="/proposal/:proposalId">
-              <Proposal />
-            </Route>
```
germanurrus, line 46: - should we leave this one and redirect to the new created route?

frontend/packages/client/src/pages/index.js
```diff
@@ -32,6 +32,9 @@ export default function AppPages() {
             <Route exact path="/community/create">
               <CommunityCreate />
             </Route>
+            <Route path="/community/:communityId/proposal/:proposalId">
+              <Proposal />
+            </Route>
             <Route path="/community/:communityId/edit" exact={true}>
```
germanurrus, line 38: we have three routes starting with the same path, I'm holding to create a separate folder with the nested route config structure, maybe if this grows we should do that

dbslone:
@germanurrus I think we should also move the `/proposal/create` endpoint to be `/community/:id/proposal/create`

germanurrus:
> @germanurrus I think we should also move the `/proposal/create` endpoint to be `/community/:id/proposal/create`

@dbslone updates where done to change create proposal url

dbslone (approved):

PR #313, opened 2022-07-27
cas 361 /results endpoint should use strategy maxWeight to calculate WIP

rvmelkonian:
* Adds logic to check and enforce `Max_weight` in `TallyVotes`
* Updates Test Suite to use the actual tally function in our controllers as opposed to duplicate logic (which kept having to be changed
* Updates Test suite to use our structs in `/models` as opposed to duplicate structs. these also kept having to be changed
* Removes unneeded Tally functions in test utils


backend/main/test_utils/strategy_utils.go
```diff
@@ -201,31 +250,35 @@ func (otu *OverflowTestUtils) GenerateSingleVoteWithNFT(
...
-	vote.Primary_account_balance = uint64(balance)
-	vote.Staking_balance = uint64(balance * 5)
-	vote.Block_height = uint64(0)
+	primary := uint64(balance)
+	staking := uint64(balance * 5)
+	blockHeight := uint64(0)
+
+	vote.PrimaryAccountBalance = &primary
+	vote.StakingBalance = &staking
+	vote.BlockHeight = &blockHeight
```
rvmelkonian, line 259: planning to clean this up so there is less code duplication on these generate functions.

backend/main/strategies_test.go
```diff
@@ -215,8 +251,15 @@ func TestStakedTokenWeightedDefaultStrategy(t *testing.T) {
 
 	t.Run("Test Tallying Results", func(t *testing.T) {
 		// Tally results
+		maxWeightDefault := float64(0.0)
+		strategyName := "staked-token-weighted-default"
+
+		s := strategyMap[strategyName]
 		proposalWithChoices := models.NewProposalResults(proposalId, choices)
-		_results := otu.TallyResultsForStakedTokenWeightedDefault(votes, proposalWithChoices)
+		_results, err := s.TallyVotes(votes, proposalWithChoices, maxWeightDefault)
```
rvmelkonian, line 259: here we use `TallyVotes` from the actual strategy itself now.

backend/main/models/proposal.go
```diff
@@ -108,7 +108,14 @@ func (p *Proposal) GetProposalById(db *s.Database) error {
 	WHERE p.id = $1
 	GROUP BY p.id`
 	sql = fmt.Sprintf(sql, computedStatusSQL)
-	return pgxscan.Get(db.Context, db.Conn, p, sql, p.ID)
+	err := pgxscan.Get(db.Context, db.Conn, p, sql, p.ID)
+
+	if p.Max_weight == nil {
```
jacksonConrad, line 113: I was thinking that if Max_weight isn't set, then there should be no Max_weight (i.e. just skip the check)

backend/main/models/vote.go
```diff
@@ -297,6 +297,16 @@ func (v *Vote) ValidateChoice(proposal Proposal) error {
 	return nil
 }
 
+func CheckForMaxWeight(maxWeight float64, balance uint64) bool {
+	floatBalance := float64(balance)
+
+	if maxWeight != 0 && floatBalance >= maxWeight {
```
jacksonConrad, line 303: oh i see, so you are saying if max_weight is 0, then skip the check

rvmelkonian: yah, arrived at this to avoid max_weight ever being `nil`. if max_weight was not set to 0 somewhere it was becoming difficult to do checks against it. I could try something like
```
if p.Max_weight != nil {
// then enforce it here 
}
```

rvmelkonian: updated with a refactor!

dbslone (approved):

PR #322, opened 2022-07-28
CAS-364: Refactor app.go to be more modular

jbluks:
**Ticket:** [CAS-364](https://linear.app/dappercollectives/issue/CAS-364/refactor-maingo-file)

**Description:**

- Based on previous PR Feedback, have updated code to focus on creating helper functions for duplicated logic.
- Did some minor file restructuring breaking routes/handlers into their own files
- Changes are not huge, but do create some more consistency and readability in the code
- Removed unnecessary comments

Note: Leaderboard tests were failing due to some other changes on the main branch. Will look at tests in separate ticket since there are bug fix tickets to be addresses around the Leaderboard specifically. Refactoring has no major impact on the Leaderboard and. all other tests pass.


rvmelkonian:
So I think this refactor doesn't achieve much. What we want to aim for is to reduce the sizes of the controller functions themselves by creating higher levels of abstraction for the logic. Here the functions have stayed the same length, but just been moved to new places, arguably creating a bit more work for the dev as they need to now remember and move between multiple files. 

I think we want to keep all the controllers in one file, but try to aim for each controller function being no longer that 10 - 15 lines (or less!). This means you will need to create new functions that hide away much of the logic being exposed already at the controller level. 

For example at lines 759 in main branch currently there are three functions :
```
// validate timestamp of request/message
	if err := a.validateTimestamp(p.Timestamp, 60); err != nil {
		log.Error().Err(err)
		respondWithError(w, http.StatusForbidden, err.Error())
		return
	}

	if err := models.EnsureRoleForCommunity(a.DB, p.Creator_addr, communityId, "author"); err != nil {
		errMsg := fmt.Sprintf("Account %s is not an author for community %d.", p.Creator_addr, p.Community_id)
		log.Error().Err(err).Msg(errMsg)
		respondWithError(w, http.StatusForbidden, errMsg)
		return
	}

	if err := a.validateSignature(p.Creator_addr, p.Timestamp, p.Composite_signatures); err != nil {
		log.Error().Err(err)
		respondWithError(w, http.StatusForbidden, err.Error())
		return
	}
```
All three can be considered part of one piece of logic; validation. So you could create a new function `ValidateProposal` inside models/proposal` and put this logic there. This would remove 18 lines from this controller function, with simply one function call ``ValidateProposal()`` If you take the same approach across all the controllers, it will reduce the length of each controller function significantly, and we can then keep all the controllers in the same file, which is desirable. 

With this type of refactoring there are no hard and fast rules about but each refactor is a case-by-case thing. 

Also when doing this avoid the temptation to just put many functions into one function, instead try and group them by logical steps in the code ... try to keep each function doing one thing that is expressed by the function name. In the example above, all three steps are required to validate the vote, they can all be grouped together, as together they achieve this one thing.  For eg. A pgx function inside validate, would be surprising as 'ValidateVote()`` doesn't suggest that the DB is being interacted with. 

Hope this makes sense!

backend/main/server/handlers/communities.go
```diff
@@ -0,0 +1,238 @@
...
+		count, _ := strconv.Atoi(r.FormValue("count"))
+		start, _ := strconv.Atoi(r.FormValue("start"))
+	
+		if count > 25 || count < 1 {
+			count = 25
+		}
+		if start < 0 {
+			start = 0
+		}
+
```
rvmelkonian, line 36: this if block is repeating logic. We can put it into its own function

backend/main/server/handlers/communities.go
```diff
@@ -0,0 +1,238 @@
...
+			return
+		}
+
+		pin, err := ipfsClient.PinJson(c)
+		if err != nil {
+			s.RespondWithError(w, http.StatusInternalServerError, "IPFS error: "+err.Error()+".")
+			return
+		}
+		c.Cid = &pin.IpfsHash
+
```
rvmelkonian, line 131: ipfs Pin logic also repeats throughout all the controllers,

backend/main/server/handlers/communities.go
```diff
@@ -0,0 +1,238 @@
...
+		// validate timestamp of request/message
+		if err := validateTimestamp(payload.Timestamp, 60, config); err != nil {
+			log.Error().Err(err)
+			s.RespondWithError(w, http.StatusForbidden, err.Error())
+			return
+		}
+
+		// validate user signature
+		if err := validateSignature(payload.Signing_addr, payload.Timestamp, payload.Composite_signatures, fa, config); err != nil {
+			s.RespondWithError(w, http.StatusForbidden, err.Error())
```
rvmelkonian, line 218: group all the valudation functions into one function something like, `ValidateUser`?

jbluks:
Got it! Thanks Rich for the feedback - will focus on function size/abstraction.

backend/main/server/handlers/communities.go
jbluks: Created helper function for this logic.

backend/main/server/handlers/communities.go
jbluks: Created helper function, although it has minimal impact on code length.

backend/main/server/handlers/communities.go
jbluks: Done. Also created a helper function that can validate user role as well, since I saw several places where that logic. was duplicated.

rvmelkonian (approved):

PR #324, opened 2022-08-01
cas 351 - Upgrade to use react form

germanurrus:
Added react-hook-form, Yup for field validation

- This PR restructures creating a community and also updating a community using the same form for both actions
- Logic for uploading avatar and banner has been placed in one component

The `community/CommunityEditorLinks` folder has the form and other components for url links 
The `community/CommunityEditorProfile` folder has the form   and other components for images and community name and other fields related to profile 

Each folder has the configuration for each form and uses `Yup` for form validation

The when creating a community the `/components/CommunityCreate/StepOne` folder has the first step for the community creation that merges the two former forms into one merging validation as well. 

 Created a `common` folder with components used with react-hook-forms

### To Test:
- Create a community, the first screen is affected by the change
- Edit a community profile. 

### Notes:
Community Creation and Proposal creation will be updated using https://react-hook-form.com/advanced-usage#WizardFormFunnel this PR servers as preparation for the future update


frontend/packages/client/src/components/Community/CommunityEditorProfile/CommunityEditorProfile.js
```diff
@@ -0,0 +1,123 @@
...
+        communityCategory: category,
+        logo: logo ? { imageUrl: logo } : undefined,
+        banner: banner ? { imageUrl: banner } : undefined,
+        communityTerms: terms,
+      },
+      resolver: yupResolver(ProfileSchema),
+    });
+
+  const logoField = useWatch({ control, name: 'logo' });
+  const bannerField = useWatch({ control, name: 'banner' });
```
germanurrus, line 48: image file has been added to form so we don't need a state to save it. We need to use useWatch to get updates and trigger renders if a new image is uploaded since we can't use register on input used by react-drop-zone

jbluks (approved):
Looks good!

PR #327, opened 2022-08-01
dev environment + results fix

rvmelkonian:
This PR fixes two issues. 

- `EnforceMaxWeight` was not comparing balance against `proposal.Max_weight` correctly 
- Local dev was not updating balances to the DB because of snapshot balance fetch override. 

Changes made
• Adds new algorithm to convert balances of `uint64` to float, with the correct decimal place shifts
* Sends back a dummy balance on `FetchBalanceFromSnapshot`, 
* Adds new function inside `token-weighted-default` and `staked-token-weighted-default`, `FetchBalanceFromSnapshot`, this cleans up `FetchBalance` quite a lot. 
* Moves `FTBalanceResponse` Struct into shared. 
* Uses that same struct for both FT and Flow Token responses from snapshotter
* Adds constructor `NewFTBalance()`, this is required if you want to update a struct by reference within another function, its fields need to be initialised. 

Results being calculated correctly. And `Max_weight` being applied correctly in my manual tests. 

Screenshot 2022-08-03 at 14 04 29

**Improvements**
There are many possible edge cases which are hard to test manually, I think unit tests would be the safest way to go here, as this functionality is so critical. 

This breaks our tests suite currently for all the `TallyVotes` methods, as `dummyBalance` is now being generated `SnapshotClient.GetAddressBalanceAtBlockHeight`

**Notes**
I fiddled with the how the results are being returned, but many variations caused the front end not to render results correctly, with the current setup it seems to work as expected.

### DIFF
--- a/backend/main/models/balance.go
+++ b/backend/main/models/balance.go
@@ -1,7 +1,6 @@
 package models
 
 import (
-	"fmt"
 	"time"
 
 	s "github.com/DapperCollectives/CAST/backend/main/shared"
@@ -34,8 +33,6 @@ func (b *Balance) GetBalanceByAddressAndBlockHeight(db *s.Database) error {
 }
 
 func (b *Balance) CreateBalance(db *s.Database) error {
-
-	fmt.Printf("Balance %+v\n", b)
 	sql := `
 	INSERT INTO balances (addr, primary_account_balance, secondary_address,
 	    secondary_account_balance, staking_balance, script_result, stakes, block_height, id)
--- a/backend/main/models/balance.go
+++ b/backend/main/models/balance.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"fmt"
 	"time"
 
 	s "github.com/DapperCollectives/CAST/backend/main/shared"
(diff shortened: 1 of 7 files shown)


backend/main/strategies/staked_token_weighted_default.go
```diff
@@ -31,16 +31,17 @@ func (s *StakedTokenWeightedDefault) FetchBalance(
...
+
+	// if err := s.SC.GetAddressBalanceAtBlockHeight(
+	// 	b.Addr,
+	// 	b.BlockHeight,
+	// 	b,
+	// 	&strategy.Contract,
+	// ); err != nil {
+	// 	log.Error().Err(err).Msg("Error fetching balance.")
+	// 	return nil, err
+	// }
```
rvmelkonian, line 44: commented this out temporarily so I don't get compiler warnings

backend/main/shared/structs.go
```diff
@@ -65,6 +67,31 @@ type MintParams struct {
...
+	SecondaryAccountBalance uint64    `json:"secondaryAccountBalance"`
+	Balance                 uint64    `json:"balance"`
+	StakingBalance          uint64    `json:"stakingBalance"`
+	ScriptResult            string    `json:"scriptResult"`
+	Stakes                  []string  `json:"stakes"`
+	BlockHeight             uint64    `json:"blockHeight"`
+	Proposal_id             int       `json:"proposal_id"`
+	NFTCount                int       `json:"nftCount"`
+	CreatedAt               time.Time `json:"createdAt"`
+}
```
rvmelkonian, line 85: this now has fields for both FT and Flow snapshot responses, so the same struct can be used for both cases.

backend/main/models/balance.go
```diff
@@ -33,17 +34,14 @@ func (b *Balance) GetBalanceByAddressAndBlockHeight(db *s.Database) error {
 }
 
 func (b *Balance) CreateBalance(db *s.Database) error {
-	// Skip for test/dev.
-	if *db.Env == "TEST" || *db.Env == "DEV" {
-		return nil
-	}
 
-	// Build SQL query to insert balances to DB
+	fmt.Printf("Balance %+v\n", b)
```
dbslone, line 38: i think this should be removed so we are not leaking data into the logs

jbluks (approved):
LGTM

dbslone (approved):

PR #328, opened 2022-08-02
FLOAT strategy 

rvmelkonian:
This PR adds all the necessary code for the Float strategy 

• adds new FlowAdapter methods to run scripts with Cadence-to-Go parsing
• adds new strategy file and updates strategyMap
• adds `owns_specific_float.cdc`
• adds custom script `get_float_ids.cdc` (we need the ids of the NFTs themselves to prevent double voting).
• refactors both NFT strategies files by adding `queryNFTs` method, this makes `FetchBalance` cleaner.  
• adds new Migration file 
• adds frontend input for `eventId` - by @jbluks 

No intergration tests have been written for this strategy yet as a lot of test_utils method need to be written with overflow in order to set up the scenario that we need to test for. I opted out of this for the time being, in favour of manual testing on mainnet. 

This has only been manually tested on mainnet. 

- All integration tests passing

If anyone wants to carry out a manual test, follow the steps:

1. Switch the dev environment to point to mainnet. Both FE and BE .envs need to be changed.
2. Claim your free FloatNFT here https://floats.city/0xfd15272b1d396f67/event/350012165
4. Create a new Community with the strategy `Float NFT`, input the following 
```
Contract Address : 0x2d4c3caffbeab845
Contract Name: FLOAT
Collection Path: FLOATCollectionPublicPath
EventId: 350012165
```
4. Cast a vote. Check for 201 Response. NB. Voter account has to be the same wallet address that claimed the FloatNFT to work.


jbluks:
Front End changes have been added - validated that event Id entered on the front-end is stored in the DB with the strategy as expected.

germanurrus (approved):
Looks good! I was able to crate a community with the Float strategy, then create a proposal, select the strategy and vote. I wasn't sure if I needed to wait on the flag `snapshotStatus: "processing"` to change it's status.

PR #332, opened 2022-08-03
CAS-319: Prevent user from deleting strategy used by active proposal

jbluks:
**Ticket:** [CAS-319](https://linear.app/dappercollectives/issue/CAS-319/can-not-access-proposals-with-deleted-voting-strategy)

**Description**

- Adds check to ensure that strategy to be deleted is not in use by a pending/active proposal
- Adds `/communities/:communityId/strategies` endpoint that returns unique strategies on any pending/active proposals
- Error message show to user when attempting to delete a strategy that is in use.

Screen Shot 2022-08-03 at 12 06 18 PM


rvmelkonian:
@jbluks just need to clean up the merge conflict here after refactor merge, the I think it's good to go

rvmelkonian (approved):

PR #334, opened 2022-08-05
CAS-365:  Validate Transaction Signature instead of User Signature (Dapper Wallet/Ledger workaround)

jacksonConrad:
## User Message Signing
On the CAST Frontend, the two message signing techniques are abstracted away into a single function: `signUserMessageByProvider` in [`contexts/Web3.js`](https://github.com/DapperCollectives/CAST/pull/334/files#diff-baedbdacd01ccf4a2352e300c4f7c4e4a22f88475838b4a1676079357567e915):
```
  /**
   *
   * @param {String} walletProvider ID of the wallet service user is connected to
   * @param {String} cadence transaction code to be authorized
   * @param {String} data transaction argument, or message data to be validated on the backend
   * @returns {[Array?, Object?]} first element is an array of composite signatures,
   * second element is voucher JSON parsed into an Object.  If function is successful, one element will be null.  If it fails, both elements will be null.
   */
  const signMessageByWalletProvider = async (walletProvider, cadence, data) => {
    try {
      let voucher;
      switch (walletProvider) {
        case 'dapper#authn':
          voucher = await signMessageVoucher(cadence, data);
          return [null, voucher];
        case 'fcl-ledger-authz':
          voucher = await signMessageVoucher(cadence, data);
          return [null, voucher];
        default:
          const _compositeSignatures = await fcl
            .currentUser()
            .signUserMessage(data);
          return [getCompositeSigs(_compositeSignatures), null];
      }
    } catch (e) {
      return [null, null];
    }
  };
  ```
  `signUserMessageByProvider` returns an array of two elements containing either an array of Composite Signatures or a signed Voucher, depending on which wallet the user is connected with.  Both are included in the request payload for the API.

### `fcl.currentUser().signUserMessage`
This is the FCL-native way of generating user signatures for an arbitrary message.  We simply pass in the hex-encoded message, the user clicks "Sign", and we receive a payload containing an array of Composite Signatures.  However, Dapper Wallet does not have an implementation for `currentUser().signUserMessage`.  For Dapper Wallet we use...

### Transaction Signatures via `signMessageVoucher`

```
  /**
   *
   * @param {String} cadence - transaction code to be authorized by user
   * @param {String} data - transaction argument, or message data
   *  we want to validate on the backend
   * @returns {Object} - voucher as JSON
   */
  const signMessageVoucher = async (cadence, data) => {
    const voucher = await fcl.serialize([
      fcl.transaction(cadence),
      fcl.args([fcl.arg(data, t.String)]),
      fcl.limit(999),
      fcl.proposer(fcl.authz),
      fcl.authorizations([fcl.authz]),
      fcl.payer(fcl.authz),
    ]);
    return JSON.parse(voucher);
  };
  ```
`fcl.serialize` prompts the user to authorize a transaction.  Upon clicking "Confirm", a transaction Voucher is created, which includes the signatures that we can validate on the backend.  The transaction is never actually run.  We send the message `data` that we want to validate on the backend as the transaction `args`.  The signatures are included in the Voucher either under `payloadSigs` or `envelopeSigs`.

## Validating Tx Signatures on CAST Backend

For API routes that require a message/signature to be submitted, the payload may optionally include a `voucher` param.  If  `voucher` is present, the API will validate the data & signatures inside the `voucher` object.  Otherwise, it will look for a `compositeSignatures` param on the payload to validate.

#### Validating the Signature
1. Rebuild the encoded transaction envelope via `func EncodeMessageFromVoucher(v *Voucher) string` in `voucher.go`
  - This function mimics the FCL SDK `encodeTransactionPayload` function [here](https://github.com/onflow/fcl-js/blob/90a80da05512c3841f4a6506845e4ee19e272776/packages/sdk/src/encode/encode.js#L6).
2. Create a `Composite_signatures` struct from the payload/envelope signatures in the Voucher via `func GetUserCompositeSignatureFromVoucher(v *Voucher) *[]CompositeSignature`
3. Validate that the signature is valid by passing the encoded message from 1) and the composite signatures from 2) to `func validateTxSignature` in `helpers.go`.
  - Note: `func (fa *FlowAdapter) ValidateSignature` now accepts an additional argument, `messageType` which can either be `USER` or `TRANSACTION`.  This allows us to use the same `validate_signature.cdc` script for both signature types.  The difference is the `domainSeparationTag` passed to `pk.verify` inside `validate_signature.cdc`.

#### Validating the Message Data

We include the message data that must be validated inside the voucher `args`.  For example, a `createProposal` request only requires a Unix timestamp to be included in the message data.  We can validate that the timestamp included in the signed message corresponds to a time within the last 5 minutes by doing:

```
timestamp := voucher.Arguments[0]["value"]
if err := h.validateTimestamp(timestamp, 5*60); err != nil {
	return err
}
```

#### Validating the Address

Lastly, we must check that the address making the API request is the same address that has both authorized the transaction and generated the signature that we validated.
```
// Validate authorizer - `addr` is the address making the API request
authorizer := voucher.Authorizers[0]
if authorizer != addr || authorizer != (*compositeSignatures)[0].Addr {
	err := errors.New("authorizer address must match voter address and envelope signer")
	log.Error().Err(err)
	return err
}
```


gregsantos:
I think `user-signature` is the better choice here universally. It offers a more streamlined approach, allowing for signed message data to be displayed, verification to be done via FCL, and no hard reliance on a specific signing implementation of payer, proposer, authorizer

frontend/packages/client/src/networks.js
```diff
@@ -13,8 +13,8 @@ const networksConfig = {
     walletDiscovery: 'https://fcl-discovery.onflow.org/testnet/authn',
     walletDiscoveryApi: 'https://fcl-discovery.onflow.org/api/testnet/authn',
```
gregsantos, line 14: will need client side for Dapper Wallet connection
`config({"discovery.wallet.method": "POP/RPC"})
`

frontend/packages/client/package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@creativebulma/bulma-tooltip": "^1.2.0",
     "@onflow/fcl": "^1.2.0",
+    "@onflow/sdk": "file:../../../../fcl-js/packages/sdk",
```
gregsantos, line 8: 🏴‍☠️

backend/main/shared/flow.go
```diff
@@ -129,6 +129,7 @@ func (fa *FlowAdapter) UserSignatureValidate(
 	}
 
 	// Load script
+	// script, err := ioutil.ReadFile("./main/cadence/scripts/validate_signature_v2.cdc")
```
gregsantos, line 132:
```suggestion
```

frontend/packages/client/src/contexts/Web3.js
```diff
@@ -71,6 +73,7 @@ export function Web3Provider({ children, network = 'testnet', ...props }) {
       'app.detail.icon': iconUrl,
       'accessNode.api': accessApi, // connect to Flow
       'discovery.wallet': walletDiscovery, // use wallets on public discovery
+      'discovery.wallet.method': 'POP/RPC',
```
jacksonConrad, line 76: How do we use this method for _only_ Dapper Wallet?

frontend/packages/client/package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@creativebulma/bulma-tooltip": "^1.2.0",
     "@onflow/fcl": "^1.2.0",
+    "@onflow/sdk": "^1.1.1",
```
germanurrus, line 8: Can you remove the package-lock.json file and get the yarn-lock.json file updated? `yarn install` on the frontend folder should do the trick

frontend/packages/client/src/contexts/Web3.js
```diff
@@ -71,6 +73,7 @@ export function Web3Provider({ children, network = 'testnet', ...props }) {
...
       'discovery.authn.endpoint': walletDiscoveryApi, // public discovery api endpoint
       'discovery.authn.include': walletDiscoveryInclude, // opt-in wallets
     });
```
germanurrus, line 79: on [this line](https://github.com/DapperCollectives/CAST/pull/334/files#diff-baedbdacd01ccf4a2352e300c4f7c4e4a22f88475838b4a1676079357567e915R113 )
you can add: 
```
 {
          f_type: 'Service',
          f_vsn: '1.0.0',
          method: 'IFRAME/RPC',
          provider: {
            icon: '/images/dapper.svg',
            name: 'Dapper Wallet',
          },
          uid: 'dapper-wallet#authn',
        },
````
to enable local dev

**Review by germanurrus — COMMENTED** [review]
Looks good! 🚀  added some comments, I was able to test setting the `.env` for front and backend with `testnet`. 
- Created a community
- created a proposal and voted on it
- joined a community 
All the above works fine 🟢

File: frontend/packages/client/src/const/index.js [hunk 7]
```diff
@@ -28,3 +28,66 @@ export const CommunityEditPageTabs = {
...
+export const UPDATE_MEMBERSHIP_TX = `
+transaction() {
+    prepare(acct: AuthAccount) {
+        // update community membership:
+        //
+        // this transaction does nothing and will not be run,
+        // it is only used to collect a signature.
+        //
+    }
+}`;
```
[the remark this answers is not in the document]
jacksonConrad: the reason for them to be broken out into separate txs is so that we can customize the messaging for each one inside of Dapper Wallet

frontend/packages/client/src/hooks/useProposal.js
```diff
@@ -67,73 +72,39 @@ export default function useProposal() {
     injectedProvider,
     proposal,
     voteData,
-    isLedger
+    walletProviderId
   ) => {
-    return isLedger
-      ? voteOnProposalLedger(injectedProvider, proposal, voteData)
-      : voteOnProposalBlocto(injectedProvider, proposal, voteData);
+    console.log('walletProvider', walletProviderId);
```
jbluks, line 77: remove console.log

frontend/packages/client/src/hooks/useCommunity.js
```diff
@@ -56,13 +60,15 @@ export default function useCommunity({
...
-          .currentUser()
-          .signUserMessage(hexTime);
 
-        const compositeSignatures = getCompositeSigs(_compositeSignatures);
+        const [compositeSignatures, voucher] =
+          await signMessageByWalletProvider(
+            user?.services[0]?.uid,
+            CREATE_COMMUNITY_TX,
+            hexTime
+          );
```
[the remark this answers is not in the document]
jacksonConrad: thats a good idea, then we can include additional data for this arg as well, and just always append `hexTime` as one of the args automatically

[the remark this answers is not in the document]
jacksonConrad: i'm putting this off for now until it is needed (i suspect it will be, because there may be another param we need to pass in) but i did consolidate all the `timestamp/hexTime` lines into a single line of code.

jacksonConrad:
One potential issue with this implementation was flagged [here](https://axiomzen.slack.com/archives/C02HM5CLCJ0/p1660779642616449?thread_ts=1660765161.888549&cid=C02HM5CLCJ0).

However, I don't think it should be a blocker for release.  We can always do a fast follow to prevent this issue.  [Linear ticket here.](https://linear.app/dappercollectives/issue/CAS-438)

--------------------------

Also, I'm setting the `discovery.wallet.method` globally for all services [here](https://github.com/DapperCollectives/CAST/pull/334#discussion_r947233378), because otherwise Dapper Wallet doesnt work properly -- ideally we would not have to set this globally, but I'm not sure how to set it specifically for a single wallet provider.  When this is set globally, it causes Blocto to open a separate pop-up window, rather than a modal.

dbslone (approved):

PR #337, opened 2022-08-08
CAS-392 - Update add address form with react hook form

germanurrus:
Depends on https://github.com/DapperCollectives/CAST/pull/324 

- Added react-hook-forms on address wallet form 
- Added address validation on chain
To Test:
- update community users on `/community/{communityId}/edit` then go to community details tab
- create a community and on second step add invalid wallet addresses


frontend/packages/client/src/components/Community/CommunityEditorDetails/FormConfig.js
```diff
@@ -0,0 +1,109 @@
...
+        excludeEmptyString: !isEditMode,
+      })
+      .test(
+        'checkValidAddressOnChain',
+        'This Address does not exist on chain',
+        async (value, context) => {
+          if (isValidFlowAddress && value !== '') {
+            let addrVal;
+            try {
+              addrVal = await isValidFlowAddress(value);
```
germanurrus, line 62: Address validation done with FCL

frontend/packages/client/src/components/Community/CommunityEditorDetails/AddressForm.js
```diff
@@ -0,0 +1,107 @@
...
+  onAddAddress,
+  addrType = 'Admins',
+  register,
+  update,
+  isValid,
+  submitComponent,
+  removeInnerForm,
+  isSubmitting = false,
+  handleSubmit,
+  showValidIcon,
```
germanurrus, line 22: I think this ended up having too many props. Another option to have less props passed could be pass the `formFieldsComponent` as a component, then the `AddressForm` component doesn't need to get that many props. The downside of it is that the parent component is aware of the composition (here https://github.com/DapperCollectives/CAST/pull/337/files#diff-bd4d5820dc252034709a92f093bb51951a9afe419194b621250729dcf34cbba2R102 we would need to pass down the `formFieldsComponent` )

frontend/packages/client/src/components/Community/CommunityEditorDetails/FormConfig.js
```diff
@@ -0,0 +1,109 @@
...
+    isEditMode
+  );
+  return yup.object().shape({
+    ...(isEditMode
+      ? { addrList: elementValidation }
+      : {
+          listAddrAdmins: elementValidation,
+          listAddrAuthors: elementValidation,
+        }),
+  });
```
germanurrus, line 107: next PR makes these fields use a variable for the name of the field

jbluks (approved):
LGTM

frontend/packages/client/src/components/Community/CommunityEditorDetails/AddressForm.js
jbluks: I think it is okay for now. Another options could be to pass a props object and use a helper to construct the props. I have done that in the past and it can be a good way to manage a bunch of props. something like `const addressProps =  getAddressProps(arg1, arg2, arg3)`

PR #344, opened 2022-08-09
refactor helpers

rvmelkonian:
no ticket for this one. 

• Continues refactor work in `helpers.go` to make functions more SRP (single-responsibility-principle) and therefore, shorter and more approachable. 
• uses `orderedParams` struct in a more efficient manner, reduces amount of params on several models methods.

### DIFF
--- a/backend/main/models/proposal.go
+++ b/backend/main/models/proposal.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/DapperCollectives/CAST/backend/main/shared"
 	s "github.com/DapperCollectives/CAST/backend/main/shared"
 	"github.com/georgysavva/scany/pgxscan"
 	"github.com/jackc/pgx/v4"
@@ -56,7 +57,12 @@ var computedStatusSQL = `
 	END as computed_status
 	`
 
-func GetProposalsForCommunity(db *s.Database, start, count int, communityId int, status string, order string) ([]*Proposal, int, error) {
+func GetProposalsForCommunity(
+	db *s.Database,
+	params shared.OrderedPageParams,
+	communityId int,
+	status string,
+) ([]*Proposal, int, error) {
 	var proposals []*Proposal
 	var err error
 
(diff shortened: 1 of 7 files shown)


backend/main/models/vote.go
```diff
@@ -69,30 +70,31 @@ const (
 
 func GetVotesForAddress(
 	db *s.Database,
-	start, count int,
+	order shared.OrderedPageParams,
```
jbluks, line 73: Could we use consistent naming for this parameter? It looks like it is either being called `order` or `params`? Let's pick one and stick to it. I am thinking `params` since the struct contains more than just the order, or if we really wanted to be more clear maybe `pageParams`

backend/main/server/controllers.go
```diff
@@ -106,12 +106,8 @@ func (a *App) getVotesForAddress(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	start, count := getPageParams(r.FormValue("start"), r.FormValue("count"), 25)
-	orderParams := shared.OrderedPageParams{
-		Start: start,
-		Count: count,
-		Order: "desc",
-	}
+	orderParams := getPageParams(r.FormValue("start"), r.FormValue("count"), 25)
```
jbluks, line 109: Could we reuse getOrderdPageParams and just pass in "desc" as the parameter? Could we get merge `getPageParams` and `getOrderedPageParams` and pass in a default when we don't have order passed through the API?

backend/main/models/proposal.go
```diff
@@ -56,7 +57,12 @@ var computedStatusSQL = `
 	END as computed_status
 	`
 
-func GetProposalsForCommunity(db *s.Database, start, count int, communityId int, status string, order string) ([]*Proposal, int, error) {
+func GetProposalsForCommunity(
+	db *s.Database,
+	params shared.OrderedPageParams,
```
jbluks, line 62: Could we have the params consistently be the last parameter as in other function calls?

jbluks (changes requested):

backend/main/models/vote.go
rvmelkonian: will use `pageParams`

rvmelkonian:
@jbluks the changed have been implemented!

jbluks (approved):
LGTM!

PR #347, opened 2022-08-09
cas 133 - Warn before exiting incomplete

germanurrus:
- Prompt for user response before leaving create proposal or community wizard
Screen Shot 2022-08-09 at 19 19 22


frontend/packages/client/src/pages/CommunityCreate.js
```diff
@@ -155,11 +158,14 @@ export default function CommunityCreate() {
   const props = {
     finalLabel: 'Publish',
     onSubmit,
-    isSubmitting: creatingCommunity && !error,
+    isSubmitting: (creatingCommunity || data) && !error,
```
germanurrus, line 161: added data to avoid this error: 
Screen Shot 2022-08-09 at 19 13 29

dbslone (approved):

PR #358, opened 2022-08-11
cas-424 adds backend support for updating community threshold

rvmelkonian:
• updates model function and struct

### DIFF
--- a/backend/main/models/community.go
+++ b/backend/main/models/community.go
@@ -35,7 +35,6 @@ type Community struct {
 	Slug                     *string     `json:"slug,omitempty"                  validate:"required"`
 	Is_featured              *bool       `json:"isFeatured,omitempty"`
 
-	//TODO we can remove this from the struct as contract data is inside Strategies now
 	Contract_name *string  `json:"contractName,omitempty"`
 	Contract_addr *string  `json:"contractAddr,omitempty"`
 	Public_path   *string  `json:"publicPath,omitempty"`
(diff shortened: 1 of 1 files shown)


dbslone (approved):

backend/main/models/community.go
```diff
@@ -73,6 +72,12 @@ type UpdateCommunityRequestPayload struct {
 	Proposal_validation      *string     `json:"proposalValidation,omitempty"`
 	Proposal_threshold       *string     `json:"proposalThreshold,omitempty"`
 
+	//TODO dup fields in Community struct, make sub struct for both to use
+	Contract_name *string  `json:"contractName,omitempty"`
+	Contract_addr *string  `json:"contractAddr,omitempty"`
+	Public_path   *string  `json:"publicPath,omitempty"`
+	Threshold     *float64 `json:"threshold,omitempty"`
+
```
germanurrus, line 80: I think the field `onlyAuthorsToSubmit` is missing

germanurrus (commented):
Looks good, left a comment

backend/main/models/community.go
germanurrus: I also see that in the db there are two fields `proposal_threshold` and `threshold`, in the community creation we are using proposalThreshold: [here](https://github.com/DapperCollectives/CAST/blob/a7c7c9fe2b32feb11649aeba97d536147f0b375f/frontend/packages/client/src/hooks/useCommunity.js#L156)
but here I see `Threshold` being used which one is the correct one?

rvmelkonian: `threshold` is being used.could you change the front end payload to always use `threshold?` on community creation?

PR #361, opened 2022-08-12
CAS-403 - Update Community-Category options text

germanurrus:
Updates category dropdown: 

Screen Shot 2022-08-12 at 11 24 44

waiting on response on CAS-403 to proceed


jbluks (approved):

dbslone (changes requested):
@germanurrus lets remove the migration files since they are empty

germanurrus:
> @germanurrus lets remove the migration files since they are empty

I wasn't sure if we were going to need them since the original request was to change the categories to:

```
X  NFT Community => => should we create this one?  
✅ Social Community
✅ Creator Community 
X  Collector Community => should this one update service?  
✅ Protocol Community 
```

The ones with the green checkbox are already existing in the db but `NFT` and `Collector` are not. Also DAO and Service are not existing in the request CAS-403. Should we create `NFT` and `Collector and keep `DAO` and `Service`? @dbslone

backend/migrations/000034_update_community_catergory.down.sql
```diff
@@ -0,0 +1,9 @@
+DELETE FROM community_types WHERE key='nft';
+DELETE FROM community_types WHERE key='collector';
+
+BEGIN;
+ALTER TYPE community_type RENAME VALUE 'collector' TO 'service';
+END TRANSACTION;
+COMMIT;
```
germanurrus, line 7: NFT is not being removed from the `community_type` I think it's not necessary to address that in the enum type

germanurrus:
@dbslone @jbluks did an update on migration file name

jbluks (approved):
LGTM

rvmelkonian (approved):
looks good and tested on my side.

PR #364, opened 2022-08-12
CAS-428 - Update react-query version

germanurrus:
Preparation Work to have latest version of react-query


frontend/packages/client/.prettierrc
```diff
@@ -7,13 +7,14 @@
...
   "importOrder": [
-    "react",
+    "^react*",
     "contexts",
     "components",
     "hooks",
     "const",
     "utils",
     "<THIRD_PARTY_MODULES>",
+    "^@*$",
```
germanurrus, line 17: added this to handle imports with '@....' I'll create a separate PR to do:  
```
    "^react*",
    "<THIRD_PARTY_MODULES>", ==> external packages after all react related imports 
    "^@*$", => imports with  @
    "^contexts*",
    "^components*",
    "^hooks*",
    "const",
    "utils",
```
this will create a lot of changes in the import order

frontend/packages/client/package.json
```diff
@@ -6,6 +6,8 @@
     "@creativebulma/bulma-tooltip": "^1.2.0",
     "@onflow/fcl": "^1.2.0",
     "@onflow/six-transfer-tokens": "0.0.8",
+    "@tanstack/react-query": "^4.1.0",
+    "@tanstack/react-query-devtools": "4.0.11-beta.0",
```
germanurrus, line 10: Had to install beta version bc 4.0.11 was not working.  Read this here: https://github.com/TanStack/query/issues/3916#issuecomment-1213172411
I think it's not an issue because we are only using this for local dev, and dev-tools are not included in prod bundle

jbluks (approved):

frontend/packages/client/src/App.js
```diff
@@ -34,7 +33,7 @@ function App() {
             </Router>
           </NotificationModalProvider>
         </Web3Provider>
-        {isLocalDevMode && <ReactQueryDevtools />}
+        <ReactQueryDevtools initialIsOpen={false} />
```
germanurrus, line 36: if NODE_ENV is not development then this component doesn't get added to the bundle and does not apear on the app.

PR #365, opened 2022-08-12
CAS-426: Simplify Leaderboard implementation

jbluks:
**Ticket:** [CAS-426](https://linear.app/dappercollectives/issue/CAS-426/update-leaderboard-implementation)

**Description:**

- The leaderboard implementation was more complicated than necessary. Realized that we don't have to track the user achievements in a separate table.
- Adds `is_early` and `is_winning` columns to the `votes` table where a vote can be indicated as an early vote and/or winning vote. It is easy to do a select and count on those columns.
- Streaks can be calculated by going through users in a community, selecting their votes and looking for streaks.
- Cancelled votes are easily excluded checking the `is_cancelled` column.


backend/main/server/controllers.go
```diff
@@ -54,7 +54,7 @@ func (a *App) getResultsForProposal(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if *proposal.Computed_status == "closed" {
-		models.AddWinningVoteAchievement(a.DB, votes, results, proposal.Community_id)
+		models.AddWinningVoteAchievement(a.DB, votes, results)
```
jacksonConrad, line 57: as we discussed, flag the proposal after all the vote stats have been updated so it doesnt attempt it more than once.

rvmelkonian (commented):
This is much cleaner! 

When I ran `testmigratedown` I got the error `column "is_early" of relation "votes" does not exist `, I think the migration `ALTER TABLE votes DROP COLUMN is_early;` is missing an `IF EXISTS`, and for `is_winning`.

And one Merge Conflict here

jbluks:
Fixed migration scripts to add IF EXISTS

rvmelkonian (approved):

PR #367, opened 2022-08-15
CAS-347: Add /add-fungible-token

jbluks:
**Ticket:** [CAS-347](https://linear.app/dappercollectives/issue/CAS-347/when-saving-strategy-validate-contract-info-is-correct-then-call-add)

**Description:**

- Adds endpoint for adding fungible token to the snapshotter when user selects a fungible token strategy
```
POST /add-fungible-token
{
  "addr": "0xabc123",
  "name": "Token",
  "path:" "tokenPath"
}
```
- API will call the `/add-fungible-token` on the snapshotter to record the token contract.


dbslone:
You should be able to test this locally if you run the snapshotter then in the .env for the API set `API_URL=PRODUCTION` and `SNAPSHOT_BASE_URL=127.0.0.1:8008`

dbslone:
> Does the snapshotter know if a token has already been added? E.g if I am adding a strategy using Flow Token, will it error, or skip?

The snapshotter assumes that the token has already been added before it tries to take the snapshot of the current blockheight. If the token has not been added then the snapshot will fail because it doesn't have the contract address.

backend/main/shared/snapshot-flow.go
```diff
@@ -207,6 +214,32 @@ func (c *SnapshotClient) GetLatestSnapshot(contract Contract) (*Snapshot, error)
...
+	}
+
+	url := fmt.Sprintf(`%s/add-fungible-token`, c.BaseURL)
+
+	payload := FungibleTokenContract{
+		ContractAddress: addr,
+		ContractName: name,
+		PublicCapabilityPath: path,
+	}
+	req, err := c.setPostRequest(url, payload)
```
rvmelkonian, line 229: inside `setPostRequest` you can call `setRequestMethod` and pass in `"POST"` as the method arg, this will remove a bit of repeating logic.

rvmelkonian (commented):
Looks Good. I'll wait till the merge conflict is fixed before I do a manual test.

backend/main/shared/snapshot-flow.go
jbluks: The reason that I have done it this way is that I need to add a body to the post request, and the setRequestMethod doesn't take a body as a parameter. Since optional arguments is a sort of go-anti pattern, I created a function that accepts a struct and sets the body. Is there a way that we can make this more generic?

rvmelkonian (approved):

PR #375, opened 2022-08-17
Validate Contract Threshold cas-414

tuffon:


backend/main/server/controllers.go
```diff
@@ -344,6 +351,13 @@ func (a *App) updateCommunity(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	//Validate Contract Thresholds
+	err = validateConractThreshold(*payload.Strategies)
```
jbluks, line 355: Spelling error - should be validateCon**t**ractThreshold

backend/main/server/controllers.go
```diff
@@ -353,6 +367,17 @@ func (a *App) updateCommunity(w http.ResponseWriter, r *http.Request) {
 	respondWithJSON(w, http.StatusOK, c)
 }
 
+func validateConractThreshold(s []models.Strategy) error {
```
jbluks, line 370: Could we put this at the bottom of the file with the helpers?

jbluks (changes requested):

jbluks:
Two small changes - but otherwise looks good!

backend/main/server/controllers.go
tuffon: nice catch

backend/main/server/controllers.go
tuffon: sure thing!

jbluks (approved):

PR #376, opened 2022-08-18
CAS-434: Fix Only Author submit not stored in DB

jbluks:
**Description:**

- The Only update by designated author field is not updated when editing a community
- Value was not being passed to updateCommunity and part of the UPDATE query.

### DIFF
--- a/backend/main/models/community.go
+++ b/backend/main/models/community.go
@@ -72,6 +72,7 @@ type UpdateCommunityRequestPayload struct {
 	Terms_and_conditions_url *string     `json:"termsAndConditionsUrl,omitempty"`
 	Proposal_validation      *string     `json:"proposalValidation,omitempty"`
 	Proposal_threshold       *string     `json:"proposalThreshold,omitempty"`
+	Only_authors_to_submit 	 *bool       `json:"onlyAuthorsToSubmit,omitempty"`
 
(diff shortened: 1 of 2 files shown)


backend/main/models/community.go
```diff
@@ -217,8 +218,9 @@ func (c *Community) UpdateCommunity(db *s.Database, p *UpdateCommunityRequestPay
 	contract_name = COALESCE($16, contract_name),
 	contract_addr = COALESCE($17, contract_addr),
 	public_path = COALESCE($18, public_path),
-	threshold = COALESCE($19, threshold)
-	WHERE id = $20
+	threshold = COALESCE($19, threshold),
```
germanurrus, line 221: to avoid confusion can we remove `threshold` ? it's been used for the creation and update requests

germanurrus (approved):

backend/main/models/community.go
jbluks: Good idea - I was actually looking at the DB in another ticket and was getting confused about that column.

jbluks: Done.

jbluks: We can do this in another ticket to clean up once we confirm that this value isn't used elsewhere unexpectedly.

dbslone (approved):

PR #380, opened 2022-08-19
CAS-427: Fix Proposal restriction using NFT Contract

jbluks:
**Description:**

- Adds a Contract Type field to the Community Proposal Page. This will allow the backend to choose the right cadence script when validating proposal creation
- Adds `Contract_type` to the `communities` table to track the contract type for proposal creation.
- If a user attempt to create a proposal without matching the specified threshold for proposal creation, they will get a 403 error.

Screen Shot 2022-08-19 at 1 53 02 PM
e

NOTE: This fix should also address: [CAS-435](https://linear.app/dappercollectives/issue/CAS-435)

### DIFF
--- a/backend/main/server/helpers.go
+++ b/backend/main/server/helpers.go
@@ -402,12 +402,12 @@ func (h *Helpers) fetchCommunity(id int) (models.Community, int, error) {
 	return community, http.StatusOK, nil
 }
 
-func (h *Helpers) createProposal(communityId int, p models.Proposal) (models.Proposal, int, error) {
+func (h *Helpers) createProposal(p models.Proposal) (models.Proposal, int, error) {
 	if err := h.validateUser(p.Creator_addr, p.Timestamp, p.Composite_signatures); err != nil {
 		return models.Proposal{}, http.StatusForbidden, err
 	}
 
-	community, httpStatus, err := h.fetchCommunity(communityId)
+	community, httpStatus, err := h.fetchCommunity(p.Community_id)
 	if err != nil {
 		return models.Proposal{}, httpStatus, err
 	}
@@ -426,7 +426,7 @@ func (h *Helpers) createProposal(communityId int, p models.Proposal) (models.Pro
 		return models.Proposal{}, http.StatusInternalServerError, err
 	}
 
-	if err := h.enforceCommunityResitrictions(community, p, strategy); err != nil {
+	if err := h.enforceCommunityRestrictions(community, p, strategy); err != nil {
 		return models.Proposal{}, http.StatusForbidden, err
 	}
 
@@ -463,7 +463,7 @@ func (h *Helpers) createProposal(communityId int, p models.Proposal) (models.Pro
 	return p, http.StatusCreated, nil
 }
 
-func (h *Helpers) enforceCommunityResitrictions(
+func (h *Helpers) enforceCommunityRestrictions(
 	c models.Community,
 	p models.Proposal,
 	s models.Strategy,
@@ -476,7 +476,19 @@ func (h *Helpers) enforceCommunityResitrictions(
 			return errors.New(errMsg)
 		}
 	} else {
-		hasBalance, err := h.processTokenThreshold(p.Creator_addr, s)
+		threshold, err := strconv.ParseFloat(*c.Proposal_threshold, 32)
+		if err != nil {
+			log.Error().Err(err).Msg("Invalid proposal threshold")
+			return errors.New("Invalid proposal threshold")
+		}
+
+		contract := shared.Contract{
+			Name: c.Contract_name,
+			Addr: c.Contract_addr,
+			Public_path: c.Public_path,
+			Threshold: &threshold,
+		}
+		hasBalance, err := h.processTokenThreshold(p.Creator_addr, contract, *c.Contract_type)
 		if err != nil {
 			errMsg := "Error processing Token Threshold."
 			log.Error().Err(err).Msg(errMsg)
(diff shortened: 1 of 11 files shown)


frontend/packages/client/src/hooks/useCommunity.js
```diff
@@ -99,8 +99,9 @@ export default function useCommunity({
           websiteUrl,
           logo,
           banner,
-          contractAdrress: contractAddr,
+          contractAddress: contractAddr,
```
germanurrus, line 102: good catch! (adding typescript in the future would be awesome!)

frontend/packages/client/src/hooks/useCommunity.js
```diff
@@ -152,6 +153,7 @@ export default function useCommunity({
               flowAddress.contractAddr
             ),
             contractName: setDefaultValue(contractN, flowAddress.contractName),
+            contractType,
```
germanurrus, line 156: I think you need to set it here on proposal creation otherwise this will be empty:
```suggestion
            contractType: setDefaultValue(contractN, 'ft'),
```
or you can set the value on the network file so it's not hardcoded here

germanurrus (changes requested):
LGTM! just one change for community creation

backend/main/server/helpers.go
```diff
@@ -914,17 +926,18 @@ func (h *Helpers) processSnapshotStatus(s *models.Strategy, p *models.Proposal)
...
 
-	if models.IsNFTStrategy(stratName) {
+	if contractType == "nft" {
 		scriptPath = "./main/cadence/scripts/get_nfts_ids.cdc"
 	} else {
 		scriptPath = "./main/cadence/scripts/get_balance.cdc"
 	}
 
-	hasBalance, err := h.A.FlowAdapter.EnforceTokenThreshold(scriptPath, address, &s.Contract)
+	fmt.Println(c.Addr)
```
dbslone, line 938: remove log

frontend/packages/client/src/hooks/useCommunity.js
jbluks: Totally! That actually cost me a bit of time because I couldn't figure out why the contract was failing.

backend/main/server/helpers.go
jbluks: Done.

frontend/packages/client/src/hooks/useCommunity.js
jbluks: Ah got it - see how that works now. Added value to networkConfig and followed pattern of other fields.

frontend/packages/client/src/components/Community/ProposalThresholdEditor/ThresholdForm.js
```diff
@@ -32,6 +34,23 @@ export default function ThresholdForm({
...
+          margin="mt-4"
+          options={[
+            {
+              label: 'NFT',
+              value: 'nft',
+            },
+            {
+              label: 'Fungible Token',
+              value: 'ft',
+            },
```
germanurrus, line 49: I think here we need to add an empty option otherwise if the user selects the field and then want to leave everything empty and check on onlyAuthors....it's not possible:

Screen Shot 2022-08-22 at 13 24 00

maybe: 

````options={[
            {
              label: 'NFT',
              value: 'nft',
            },
            {
              label: 'Fungible Token',
              value: 'ft',
            },
            {
              label: '' or 'None',
              value: '',
            },

**Review by germanurrus — APPROVED** [review]
LGTM!

File: frontend/packages/client/src/components/Community/ProposalThresholdEditor/ThresholdForm.js (hunk 4 above)
Comment (jbluks) [reply frontend/packages/client/src/components/Community/ProposalThresholdEditor/ThresholdForm.js:53]: Added. I think Default makes sense because it just means use whatever the default is and the user doesn't care about that value.

**Review by germanurrus — APPROVED** [review]
LGTM!

File: backend/main/server/helpers.go [hunk 5]
```diff
@@ -476,7 +476,19 @@ func (h *Helpers) enforceCommunityResitrictions(
 			return errors.New(errMsg)
 		}
 	} else {
-		hasBalance, err := h.processTokenThreshold(p.Creator_addr, s)
+		threshold, err := strconv.ParseFloat(*c.Proposal_threshold, 32)
+		if err != nil {
```
rvmelkonian, line 480: shouldn't this be 64 for `bitSize` param? `shared.Contract.Threshold` is of type `float64` ? Or some specific reason here why it's set to 32?

backend/migrations/000034_alter_communities_table.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE communities DROP COLUMN contract_type;
```
rvmelkonian, line 1: does the frontend ensure this value will never be `nil`? Always either nft or ft type?

rvmelkonian (commented):
looks good, just two small questions.

backend/main/server/helpers.go
jbluks: Good catch - it is probably a copy/paste error when I was looking up parsing floats from strings.

backend/migrations/000034_alter_communities_table.down.sql
jbluks: Yep - German commented on that. Default falls back to ft.

backend/main/server/helpers.go
rvmelkonian: another PR fixes this

PR #381, opened 2022-08-22
cas-407 script for make proposal

rvmelkonian:
• Adds new Make command `make proposal` which will create four proposals for a community created by the service account, and update the `flow_snapshot` DB. 

• Removes old commented out code.

• Adds new test. This is a workaround, it seemed easier to use our test_utils code to do this, even though its not really an integration test. 

This means we can quickly generate multiple proposals without having to go through the UI flow multiple times. 

It's actually a bit tricky passing flag args into unit tests because of the way the go test binary works. So for now I hard coded the numProposals as 4.


jbluks (approved):
LGTM

PR #403, opened 2022-08-24
CAS-442 Implement Hotjar

germanurrus:
- Installs [@hotjar/browser ](https://www.npmjs.com/package/@hotjar/browser) 
- Updates to pipeline


frontend/Dockerfile
```diff
@@ -18,6 +18,7 @@ ENV REACT_APP_BACK_END_SERVER_API "https://api.cast.fyi"
 ENV REACT_APP_IPFS_GATEWAY "https://dappercollectives.mypinata.cloud/ipfs/"
 ENV REACT_APP_FLOW_ENV "mainnet"
 ENV REACT_APP_TX_OPTIONS_ADDRS "0xe0de919ed4ebeee4,0x7f81b82fa0e59b17"
+ENV REACT_APP_HOTJAR_SITE_ID=REPLACE_HOTJAR_SITE_ID
```
germanurrus, line 21: this is actually not making the variable hidden but will be inside the bundle, and not that visible as if we add it in the lib call. Otherwise this will require to use something like this  [runtime-env-cra ](https://www.npmjs.com/package/runtime-env-cra)

dbslone: @germanurrus I think we should be able to use github secrets and modify the github action to insert the value. @fersan1985 should be able to help you with this

fersan1985: Sure, that's what we did [Here](https://github.com/DapperCollectives/CAST/pull/403/commits/bc407eb6531c78d08a809f15c7c9f1ea27f52924) @dbslone

dbslone (approved):

PR #404, opened 2022-08-24
cas 425 `make float` so that we can test float on local dev

rvmelkonian:
This PR adds a new make command `make float` this adds one FLOAT nft into the admin local emulator wallet address `0xf8d6e0586b0a20c7`

If we need to test float functionality locally in the emulator its now possible by running this command. 

N.B if you've already ran `make nft` you'll have to quit the emulator then run `make float` as both commands deploy some of the same dependancy contracts to the same address.  `make float` internally calls `make nft`


jbluks (approved):
LGTM

PR #406, opened 2022-08-24
CAS-457 - Make wallet method to be configured by service information

germanurrus:
- This is a preparation change for when blocto switches on mainnet from `'IFRAME/RPC'`  to `'POP/RPC'` . This change is on testnet. 
To test the PR, on `packages/client/.env` update 
from 
`REACT_APP_FLOW_ENV=emulator` 
to 
`REACT_APP_FLOW_ENV=testnet` 
and run the frontend 

On mainent blocto is still using `'IFRAME/RPC'` so when setting `REACT_APP_FLOW_ENV=mainnet`, no change is observed. This is why on PROD connecting wallet with blocto still works.


frontend/packages/client/src/components/WalletConnectModal.js
```diff
@@ -32,12 +32,8 @@ export default function WalletConnectModal({
...
         services.map((service) => ({
           connectToService: () => {
-            if (service.uid === 'dapper-wallet#authn') {
-              // dapper wallet uses POP/RPC
-              fcl.config().put('discovery.wallet.method', 'POP/RPC');
-            } else {
-              // other wallets are fine with IFRAME/RPC
-              fcl.config().put('discovery.wallet.method', 'IFRAME/RPC');
+            if (service.uid !== 'Lilico') {
+              fcl.config().put('discovery.wallet.method', service.method);
```
germanurrus, line 36: Blocto is deprecating IFRAME/RPC and switching to POP/RPC, so far it was been done on mainnet, this change will enable to use service method dynamically until mainnet and testnet are the same for blocto, then at some point all services will use 'POP/RPC'  excluding Lilico

rvmelkonian (approved):

PR #408, opened 2022-08-25
cas 194 - remove `threshold` at community level 

rvmelkonian:
* Removes `threshold` from DB, `Community` Struct and community model methods. 
* Checks Tests

`Community` Struct and column in DB had two fields `proposal_threshold` and `threshold`, of different types. Also the `Contract` struct has also a `threshold` field. 

The front end only uses `proposal_threshold` field as payload, so this dup `threshold` field was creating confusion. `enforceCommunityRestrictions` parses `proposal_threshold` and does nothing with `Community.threshold`. It does use `Contract.threshold`, as intended.

*N.B*
I'm Unable to test through the front end at the moment due to this error I'm getting when I run `yarn start``Error: Cannot find module @babel/helper-get-function-arity`, I have ran `yarn install` inside `/frontend`.


germanurrus (approved):
LGTM!

rvmelkonian:
don't merge yet, checking another case with this.

jbluks (approved):
LGTM

PR #409, opened 2022-08-25
CAS-433: Move Threshold and Weight to Proposals

jbluks:
Description:

- Moves Threshold and Weight inputs to Proposals Page from Community Page

Screen Shot 2022-08-25 at 9 30 07 AM

### DIFF
--- a/frontend/packages/client/src/App.sass
+++ b/frontend/packages/client/src/App.sass
@@ -544,6 +544,16 @@ span[data-tooltip]
 	-o-animation: fadein 0.5s
 	animation: fadein 0.5s
 
+.fade-out
+	-webkit-transition: opacity 0.3s ease-out
+	-moz-transition: opacity 0.3s ease-out
+	-ms-transition: opacity 0.3s ease-out
+	-o-transition: opacity 0.3s ease-out
+	opacity: 0
+
+.section.section-small
+	padding-top: 2rem !important	
+
 .word-break
 	word-break: break-word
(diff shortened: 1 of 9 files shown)


frontend/packages/client/src/components/ProposalCreate/StepOne/index.js
```diff
@@ -460,6 +460,32 @@ const StepOne = ({
...
+                className="rounded-sm border-light p-3 column is-full"
+                value={stepData?.maxWeight}
+                onChange={(event) =>
+                  onDataChange({
+                    maxWeight: event.target.value,
+                  })
+                }
+              />
+            </>
+          )}
```
germanurrus, line 488: these two fields are not being validated, this means that if the user does not complete them then it will default to the strategy entered values right?

germanurrus (approved):
LGTM! Tested locally and saves both fields on DB correctly

frontend/packages/client/src/components/ProposalCreate/StepOne/index.js
germanurrus: yes it's done here: https://github.com/DapperCollectives/CAST/pull/409/files#diff-5d87eb621c5f34cd64b6ae816ecd3b248a750bbff56f7e17c81780b9b52e7808R470

jbluks: Correct!

PR #419, opened 2022-08-29
cas 422 - Local dev to run script on snapshotter bypass

rvmelkonian:
Now in local dev, the backend checks the users wallet for actual emulator `FlowToken` and returns the balance.

* adds `GetFlowBalance` if `c.bypass == true`
* adds `-ipfs-override` flag. 

Prior to this PR, when the snapshotter being overridden for local dev, a `dummyBalance` Struct was being returned. This meant that the `collectionPath` would also always be correct and not nil in local dev, even when a wrong one would be entered through the Front End. The expected Error was not returning. 

This lead to some confusion in local testing.  Now we have a closer environment to prod for testing scenarios locally. 

Also now if a developer is having problems hitting ipfs locally they can run the server with `make macdev -ipfsoverride=true` to override ipfs call and return a dummy hash.


jbluks (approved):
LGTM

PR #420, opened 2022-08-29
CAS-467 - Add tooltip for DW Safari Users

germanurrus:
Added tooltip:

Screen Shot 2022-08-29 at 13 16 00
Screen Shot 2022-08-29 at 13 16 11


frontend/packages/client/src/hooks/useLocalStorage.js
```diff
@@ -0,0 +1,39 @@
...
+      // Save to local storage
+      if (typeof window !== 'undefined') {
+        window.localStorage.setItem(key, JSON.stringify(valueToStore));
+      }
+    } catch (error) {
+      console.log(error);
+    }
+  };
+  return [storedValue, setValue];
+}
```
germanurrus, line 39: we can use this hook to persist in local storage any value we want

frontend/packages/client/src/pages/Home.js
```diff
@@ -39,8 +47,30 @@ export default function HomePage() {
     'pt-6-mobile pt-6-tablet pt-9-desktop': isMyCommunitiesVisible,
   });
 
+  const browserName = useBrowserName();
+  const [showToolTip, setValue] = useLocalStorage('dw-safary-tooltip', null);
+
+  useEffect(() => {
+    const isSafary = 'Apple Safari' === browserName;
```
jbluks, line 54: Small thing - it should be "Safari" with an "i"

jbluks (changes requested):
One small typo that should be fixed.

jbluks (approved):

PR #422, opened 2022-08-29
CAS-443: DPS Archive Node Integration

jacksonConrad:
- Replace `shared/snapshot-flow.go` with `shared/dps.go`
- Remove all code referencing Snapshots/SnapshotBlocks
- Remove FE code for adding a fungible token to the snapshotter
- Updates all Strategy code to use DPS to fetch token balance at block height


backend/main/server/controllers.go
```diff
@@ -784,6 +710,29 @@ func (a *App) removeUserRole(w http.ResponseWriter, r *http.Request) {
...
+	log.Info().Msgf("getting balance for %s at blockheight %d", addr, blockheight)
+
+	name := "StarlyToken"
+	path := "starlyTokenBalance"
+	contractAddr := "0x142fa6570b62fd97"
+	contract := shared.Contract{Name: &name, Addr: &contractAddr, Public_path: &path}
+	// result := a.DpsAdapter.GetTokenBalanceAtBlockheight(addr, uint64(blockheight), &contract)
+	// name := "FlowToken"
+	// contractAddr := "0x142fa6570b62fd97"
+	// contract := shared.Contract{Name: &name, Addr: &contractAddr}
```
jacksonConrad, line 726: Gotta fix this route

backend/main/strategies/staked_token_weighted_default.go
```diff
@@ -53,29 +53,30 @@ func (s *StakedTokenWeightedDefault) FetchBalanceFromSnapshot(
...
-		if err := s.SC.GetAddressBalanceAtBlockHeight(
-			b.Addr,
-			b.BlockHeight,
-			ftBalance,
-			&strategy.Contract,
-		); err != nil {
-			log.Error().Err(err).Msg("Error fetching balance from snapshot client")
-			return err
-		}
+		// if err := s.DPS.GetAddressBalanceAtBlockHeight()
```
jbluks, line 56: Why are these lines commented out? Should they be deleted?

backend/main/strategies/staked_token_weighted_default.go
```diff
@@ -53,29 +53,30 @@ func (s *StakedTokenWeightedDefault) FetchBalanceFromSnapshot(
...
-		if err := s.SC.GetAddressBalanceAtBlockHeight(
-			b.Addr,
-			b.BlockHeight,
-			ftBalance,
-			&strategy.Contract,
-		); err != nil {
-			log.Error().Err(err).Msg("Error fetching balance.")
-			return err
-		}
+		// if err := s.SC.GetAddressBalanceAtBlockHeight(
```
jbluks, line 71: Same here

jbluks (changes requested):
One question about some commented out code.

backend/main/shared/dps.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	// Initialize the DPS API client.
+	codec := zbor.NewCodec()
+
+	conn, err := grpc.DialContext(
+		adapter.Context,
+		adapter.URL,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.FailOnNonTempDialError(true), // DialContext will fail immediately if passed bad connection string
```
rvmelkonian, line 73: do we require the `insecure` package for some special reason? Seems as if `credentials` grpc could be used to pass in credentials..

backend/main/shared/dps.go
```diff
@@ -0,0 +1,213 @@
...
+		log.Error().Str("dps", adapter.URL).Err(err).Msg("could not dial API host")
+		panic("error initializing DPS gRPC connection")
+	}
+
+	client := dpsApi.NewAPIClient(conn)
+	index := dpsApi.IndexFromAPI(client, codec)
+
+	cacheSize := uint64(100_000_000)
+	_invoker, err := invoker.New(index, invoker.WithCacheSize(cacheSize))
+	adapter.Invoker = _invoker
```
rvmelkonian, line 87: guessing the value for `cacheSize` has come from the dps repo?

backend/main/shared/dps.go
```diff
@@ -0,0 +1,213 @@
...
+	_script, err := ioutil.ReadFile("./main/cadence/scripts/get_flow_token_balances.cdc")
+
+	if err != nil {
+		log.Error().Err(err).Msgf("Error reading cadence script file.")
+		return nil, err
+	}
+
+	// Replace script args
+	script := dps.Config.InsertCoreContractAddresses(string(_script))
+
```
rvmelkonian, line 116: the clean up here seems good with `InsertCoreContractAddresses`!

rvmelkonian: It would be good if `InsertCoreContractAddresses` could return an error, incase one of those values is removed from the flow.json. Could be a nasty thing to debug. ..
`ReplaceAllString` doesn't natively return an error, but maybe there is a way to error handle this logic?

backend/main/shared/dps.go
```diff
@@ -0,0 +1,213 @@
...
+	if dps.bypass() {
+		log.Info().Msgf("overriding dps service for emulator")
+		return &DummyBalance, nil
+	}
+
+	var err error
+	var result cadence.Value
+
+	tokenBalanceResponse := FTBalanceResponse{}
+	tokenBalanceResponse.NewFTBalance()
```
rvmelkonian, line 135: maybe this is for a new separate PR.. but `NewFTBalance` will return hardcoded dummy balances if running on the emulator, but I think now with DPS we want to ping DPS on local dev too right?

backend/main/shared/dps.go
```diff
@@ -0,0 +1,213 @@
...
+
+		return &tokenBalanceResponse, nil
+	} else {
+		result, err = dps.GetTokenBalanceAtBlockheightScript(addr, blockheight, c)
+		if err != nil {
+			return nil, err
+		}
+		fields := result.(cadence.Struct).Fields
+		tokenBalanceResponse.Balance = uint64(fields[1].ToGoValue().(uint64))
+
```
rvmelkonian, line 157: lets hope the index number of these values returned doesn't change ... seems strange to me that Non-Flow FT's are at index 0 and `FlowToken` `PrimaryAccountBalance` is at index 1.`

rvmelkonian: What happens when an account has no balance? does it return with `0` or `nil`?

backend/main/shared/dps.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func (dps *DpsAdapter) GetTokenBalanceAtBlockheightScript(addr string, blockheight uint64, c *Contract) (cadence.Value, error) {
+	var args []cadence.Value
+
+	// Prepare Args
+	flowAddress := flow.HexToAddress(addr)
+	cadenceAddress := cadence.NewAddress(flowAddress)
+
+	args = append(args, cadenceAddress)
```
rvmelkonian, line 169: these three lines repeat `GetFlowBalanceAtBlockheightScript`. Separate func something like `preapareArgs` returning `args`?

backend/main/shared/dps.go
```diff
@@ -0,0 +1,213 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"strings"
+
+	"github.com/onflow/cadence"
+	dpsApi "github.com/onflow/flow-dps/api/dps"
+	"github.com/onflow/flow-dps/codec/zbor"
+	"github.com/onflow/flow-dps/service/invoker"
+	"github.com/onflow/flow-go-sdk"
+	_flow "github.com/onflow/flow-go/model/flow"
```
rvmelkonian, line 14: when I saw `_flow` it confused me as I thought it was using underscores for dup param names and types like in cadence. How about naming it `flowGo`? I feel like that way I would straightaway know it means the `flow-go` sdk

backend/main/strategies/staked_token_weighted_default.go
jacksonConrad: oh whoops! i think i just forgot to finish implementing this strategy

backend/main/shared/dps.go
jacksonConrad: yes, although i've removed the optional param because its already set as this (~100 MB) by default

backend/main/shared/dps.go
jacksonConrad: Yeah I'm not sure... I was definitely debugging it when I was writing it.  If a placeholder doesnt get replaced, or is replaced incorrectly, I was getting a script contract error from the emulator.  Its not the _worst_ thing to debug.  I'm just not sure what the right way to detect an error is in this case.

backend/main/shared/dps.go
jacksonConrad: `0` -- the `AccountInfo` struct initializes all values and then updates relevant ones after it fetches:

```
pub struct AccountInfo {
    pub(set) var primaryAcctBalance: UFix64
    pub(set) var secondaryAddress: Address?
    pub(set) var secondaryAcctBalance: UFix64
    pub(set) var stakedBalance: UFix64
    pub(set) var hasVault: Bool
    pub(set) var stakes: String

    init() {
        self.primaryAcctBalance = 0.0 as UFix64
        self.secondaryAddress = nil
        self.secondaryAcctBalance = 0.0 as UFix64
        self.stakedBalance = 0.0 as UFix64
        self.hasVault = true
        self.stakes = ""
    }
}
```

backend/main/shared/dps.go
jacksonConrad: yeah i'm conflicted on this. i think these 3 lines are only re-used once... but its very specific to the scripts. it would only work for scripts accepting a single address arg.  i don't think its worth refactoring at this point but if we end up adding additional scripts in the future, it might be useful to have an ArgBuilder or something like that.

backend/main/strategies/staked_token_weighted_default.go
jacksonConrad: fixed!

backend/main/shared/dps.go
rvmelkonian: ah ok in that case, if we're getting error returned I think its ok.

backend/main/shared/dps.go
rvmelkonian: thanks for sharing this info. 0 is good.

backend/main/shared/dps.go
rvmelkonian: ok got it

backend/main/shared/dps.go
rvmelkonian: Sounds good!

jbluks:
@jacksonConrad Now that we are going to be using release branches, can you retarget this PR to branch `release-v1.5.0`?

rvmelkonian:
does this need to be updates with new vote choice keys?

jacksonConrad:
Cleanest way to resolve these conflicts may just be to rebase @jbluks RCV branch with these changes.

jbluks:
This PR is handled here: https://github.com/DapperCollectives/CAST/pull/571

PR #431, opened 2022-09-07
community card + join button updates

mannynotfound:
- make community avatars fully circular
- make community cards more rounded
- update join community button to an eye icon with several hover states / tooltip
- refactor join community button to take a `size` prop where `small` is just the icon and `large` includes the icon + text
- hide the eye icon in your already joined communities

image
image

image
image


jbluks (approved):
LGTM

PR #432, opened 2022-09-08
update homepage styles

mannynotfound:
- header styles / copy updated
- bg added to homepage splash 
- update button styles
- add more top padding to footer

image


frontend/packages/client/src/App.sass
```diff
@@ -9,7 +9,7 @@ $red: #B20D0D
 $error-red: #B20D0D
 $yellow: #FBD84D
 $orange: #F4AF4A
-$light-gray: #DCDCDC
+$light-gray: #F9F9F9
```
mannynotfound, line 12: this was only being used for a non-existent selector so I deleted rules and updated this variable for the light gray in latest designs

frontend/packages/client/src/App.sass
```diff
@@ -715,9 +719,6 @@ span[data-tooltip]
 	.header-text
 		font-size: 2.5rem
 		line-height: 3rem !important
-.beta-label
-	font-size: 10px
-	font-weight: 700
```
mannynotfound, line 720: deleted this in favor of `has-weight-bold-text` + a `smallest-text` class since this was only used once or twice

frontend/packages/client/src/components/Header.js
```diff
@@ -18,7 +19,8 @@ function Header(props) {
   return (
     <>
       <header
-        className={`header ${props.location.pathname} has-background-white is-block navbar is-fixed-top`}
+        id="navbar"
```
mannynotfound, line 22: added this id as the `.header` class felt too generic for the specific style rule of the 4px yellow border

frontend/packages/client/src/components/Label.js
```diff
@@ -3,7 +3,7 @@ import classnames from 'classnames';
 
 export default function Label({ labelText = '', classNames } = {}) {
   const className = classnames(
-    'has-background-black has-text-white rounded-sm px-3 py-2 mr-2',
+    'has-background-black has-text-white rounded-lg px-3 py-2',
```
mannynotfound, line 6: removed `mr-2` from here as it wasn't needed in every instance and should be defined in the layout that includes `Label`

frontend/packages/client/src/pages/Home.js
```diff
@@ -63,7 +63,7 @@ export default function HomePage() {
   });
 
   return (
-    <section className={classNames}>
+    <>
```
germanurrus, line 66: This is making the tooltip to be next to the navbar: 
Screen Shot 2022-09-09 at 11 31 24

Maybe adding some margin within the TooltipMessage component would work? This component is only to be shown on safari

germanurrus (commented):
Looks really good! I left some comments

frontend/packages/client/src/pages/Home.js
mannynotfound: ended up removing `message-container` from `TooltipMessage` since it was only controlling margin bottom. Now it takes a `className` so we just give it `my-6` for even top/bottom margin

germanurrus (approved):
LGTM!

PR #433, opened 2022-09-08
cas 484 - Create search endpoint for Communities

rvmelkonian:
• creates new endpoint `/communities/search/{query}`
• creates new controller, helper and models method. 
• adds new migrations that install the extensions and indexing needed

I used the examples laid out here [https://arctype.com/blog/postgres-fuzzy-search/#phonetic-match]( ) I had mixed results with all of them, the query that seemed to perform best is the one in this PR: 
`SELECT name FROM communities WHERE SIMILARITY(name, $1) > 0.1`,

For the front end we might need to use url encoding on the param, for example if search query has a space in it like `nba topshots`

Just tested with Postman. No integration tests written for this feature


jbluks (approved):
LGTM

PR #435, opened 2022-09-08
update global styles / classes

mannynotfound:
- remove Arimo/Roboto/Helvetica, make DM Sans the primary font for everything
- refactor color variables, remove unused variables and add "success" and "danger" color for success/error messages
- switch any `gray` classes to `grey` as Bulma uses `grey` convention, refactor grey classes to be simpler
- update some of our grey shades
- remove `$red` and `.has-text-red` as we were using it exclusively for error messages, which in that case we should actually use Bulma's `$danger` and `has-text-danger`

image

### DIFF
--- a/frontend/packages/client/src/App.sass
+++ b/frontend/packages/client/src/App.sass
@@ -1,23 +1,17 @@
-@import url('https://fonts.googleapis.com/css?family=Arimo:regular,bold,italic&subset=latin,latin-ext')
-@import url('https://fonts.googleapis.com/css?family=Roboto+Mono:regular,bold,italic&subset=latin,latin-ext')
+@import url('https://fonts.googleapis.com/css?family=DM+Sans:regular,bold,italic&subset=latin,latin-ext')
 @charset "utf-8"
 
 // Set brand colors
-$green: #0E5B20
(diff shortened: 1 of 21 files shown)


frontend/packages/client/src/App.sass
```diff
@@ -1,23 +1,17 @@
...
+$grey: #636363
 
 // Update Bulma's global variables
-$family-primary: "Arimo"
-$family-secondary: "Arimo"
-$family-code: "Roboto Mono"
-$family-monospace: "Roboto Mono"
-$title-family: $family-secondary
+$family-sans-serif: "DM Sans"
+$family-monospace: "DM Sans"
```
germanurrus, line 14: I think this class also needs to be updated:
```
.image-caption-draft-js
	.public-DraftStyleDefault-block.public-DraftStyleDefault-ltr
		margin: 0px !important
		font-family: 'Arimo'
		font-style: normal
		font-weight: 400
		font-size: 12px
		line-height: 140%
		color: #757575
`````
maybe remove the `font-family: 'Arimo'`

germanurrus (approved):
LGTM!

PR #446, opened 2022-09-13
CAS-501 Proposal view results

germanurrus:


jbluks:
@germanurrus Now that we are going to be using Release branches going forward, can you retarget this to release branch `release-v1.5.0`

PR #449, opened 2022-09-13
add shared-components package

mannynotfound:
Adds `frontend/packages/shared-components` as the place where components that are re-used many times to live. Currently this includes the following components, listed in order of the amount of files they appear in:

Svg - 31
Loader - 13
FadeIn - 13
WrapperResponsive - 12
ActionButton - 11
WalletConnect - 6
Input - 5
Dropdown - 5
Error - 5
Label - 3
AddButton - 3
Tooltip - 2
Title - 2
Tablink - 2
StatusLabel - 2

This PR only handles the `Svg` component, turning into a high order component that takes a `name` prop and lazy loads the corresponding Svg icon, based on the work of Linxiao in https://github.com/DapperCollectives/VESSEL/pull/102

Additionally, this PR adds [storybook](https://storybook.js.org/docs/react/why-storybook) to the `shared-components` repo so we can easily test, document, and iterate on shared components there. In follow up PRs I will move the above components to `shared-components`, add global styles and any necessary storybook addons.

image


frontend/package.json
```diff
@@ -10,10 +10,10 @@
   "workspaces": {
     "packages": [
       "packages/*"
-    ],
-    "nohoist": [
-      "**"
```
mannynotfound, line 15: `nohoist: **` was removed since we need packages to share the same `node_modules` for things like `react` and `react-dom`

frontend/package.json
```diff
@@ -10,10 +10,10 @@
...
     ]
   },
-  "dependencies": {}
+  "dependencies": {},
+  "resolutions": {
+    "babel-loader": "8.1.0"
+  }
```
mannynotfound, line 18: this `babel-loader` resolution was added to resolve the following problem: https://github.com/storybookjs/storybook/issues/5183

frontend/packages/client/craco.config.js
```diff
@@ -0,0 +1,24 @@
...
+        const include = Array.isArray(match.loader.include)
+          ? match.loader.include
+          : [match.loader.include];
+
+        match.loader.include = include.concat(packages);
+      }
+      return webpackConfig;
+    },
+  },
+};
```
mannynotfound, line 24: [craco](https://www.npmjs.com/package/@craco/craco) is used so we can load things outside of `packages/client` with our create-react-app babel config

frontend/packages/client/jsconfig.json
```diff
@@ -1,7 +1,10 @@
 {
   "compilerOptions": {
-    "baseUrl": "src"
+    "baseUrl": "src",
+    "jsx": "react-jsx"
```
mannynotfound, line 4: add `jsx: react-jsx` so we dont have to `import React from 'react'` everywhere just to use JSX

frontend/packages/client/package.json
```diff
@@ -3,14 +3,15 @@
...
     "@creativebulma/bulma-tooltip": "^1.2.0",
     "@hookform/resolvers": "^2.9.7",
     "@hotjar/browser": "^1.0.6",
     "@onflow/fcl": "^1.2.0",
     "@onflow/sdk": "^1.1.1",
     "@onflow/six-transfer-tokens": "0.0.8",
-    "@tanstack/react-query": "^4.1.0",
-    "@tanstack/react-query-devtools": "4.0.11-beta.0",
+    "@tanstack/react-query": "4.3.3",
+    "@tanstack/react-query-devtools": "4.3.3",
```
mannynotfound, line 14: had to specify exact versions because I was running into this bug that was seemingly introduced yesterday: https://github.com/TanStack/query/issues/4155

frontend/packages/shared-components/src/Svg.stories.js
```diff
@@ -0,0 +1,50 @@
...
+        'Instagram',
+        'InvalidCheckMark',
+        'LinkOut',
+        'Logo',
+        'Plus',
+        'Star',
+        'Twitter',
+        'Upload',
+        'ValidCheckMark',
+        'Website',
```
mannynotfound, line 40: unfortunately have to maintain a list of options for the storybook as there wasnt a pretty way to dynamically get all the filenames of the Svgs the way we do with dynamic `import`

frontend/packages/client/jsconfig.json
germanurrus, line 4: there are still some `import React from 'react':` in the components code, but can safely be removed later on a second PR to avoid increasing the size if this one (It's not the purpose of the PR)
Screen Shot 2022-09-14 at 11 35 03

frontend/packages/shared-components/src/Svg.stories.js
germanurrus: I wonder if there's a way to show all icons in one component, so we can easily view them/search for them

frontend/packages/shared-components/src/Svg.stories.js
```diff
@@ -0,0 +1,50 @@
...
+      ],
+    },
+  },
+};
+
+const Template = (args) => <Svg {...args} />;
+export const Controls = Template.bind({});
+Controls.args = {
+  name: 'Eye',
+};
```
germanurrus, line 50: Should we have this file inside the Svg folder? So each shared component has it's own `.stories` file and on a high level we have folders only. Also for updating the .stories file and at the same time the component structure, might be easier and closer

germanurrus (approved):
Looks really good! Tested locally and works fine 👍

PR #451, opened 2022-09-13
Fix provider icon on wallet list

chasefleming:
Discovery had an issue that was returning relative paths. Now you can just use provider.icon as the src.

Addresses: https://github.com/DapperCollectives/CAST/issues/452


frontend/packages/client/src/components/WalletConnectModal.js
```diff
@@ -5,13 +5,6 @@ import classnames from 'classnames';
 import sortBy from 'lodash/sortBy';
 import { ArrowRight, Close } from './Svg';
 
-const getWalletIcon = (provider) => {
-  if (provider?.name === 'Lilico') {
-    return 'https://raw.githubusercontent.com/Outblock/Lilico-Web/main/asset/logo-dis.png';
-  }
-  return `https://fcl-discovery.onflow.org${provider.icon}`;
-};
-
```
germanurrus, line 14: I think we still need this function when `REACT_APP_FLOW_ENV=emulator`  because icons are not returning absolute paths on `provider.icon`

frontend/packages/client/src/components/WalletConnectModal.js
```diff
@@ -40,7 +33,7 @@ export default function WalletConnectModal({
             );
             closeModal();
           },
-          icon: getWalletIcon(service.provider),
```
germanurrus, line 43: maybe doing this:
```
 icon: IS_LOCAL_DEV
            ? getWalletIcon(service.provider)
            : service.provider.icon,

chasefleming: Ok updated @germanurrus . Just out of curiosity, why is it like that on local dev instead of just using the Discovery API with testnet services?

germanurrus (approved):

frontend/packages/client/src/components/WalletConnectModal.js
germanurrus: We are developing in local dev using emulator for connecting wallets, this way we can deploy contracts to a controlled env that gives us more flexibility to develop

PR #453, opened 2022-09-14
removed return error on 0 threshold logic

rvmelkonian:
•Removes controller level logic which returned an error if proposal threshold was set to 0

**Notes**
This is a quick fix to allow for community creation when `onlyAuthorsToSubmitProposals` is true, as the old code would make the backend throw an error here. 

Proper fix is, to write logic at the helper level that first: 
checks if `onlyAuthorsToSubmitProposals` is `true`. 
If yes, ensure that `proposalThreshold >= 1`


germanurrus (approved):
Tested locally and works fine

backend/main/server/controllers.go
```diff
@@ -359,14 +359,6 @@ func (a *App) createCommunity(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
-	propThreshold, err := strconv.ParseFloat(*payload.Proposal_threshold, 64)
-	if err != nil {
-		respondWithError(w, http.StatusBadRequest, "Error Converting Proposal Threshold to Float.")
-	}
```
germanurrus, line 365: maybe we can keep this logic

jbluks:
We should maybe ask @tuffon about this, since these changes revert this commit: https://github.com/DapperCollectives/CAST/commit/5fe47b3bc74562ab5c8a11f07eff3160aa07da0c

rvmelkonian:
Here's the original ticket https://linear.app/dappercollectives/issue/CAS-414. @jbluks yeah it was being set as 0 before if `onlyAuthorsToSubmitProposals` was set to true, so I would say valid for this case.

PR #458, opened 2022-09-14
add shared-components to Dockerfiles

mannynotfound:
copy the `shared-components` package as well when we do docker builds


mannynotfound:
beaten to the punch https://github.com/DapperCollectives/CAST/pull/457

PR #460, opened 2022-09-14
Fix COPY path to shared-components in docker

jbluks:


mannynotfound (approved):

frontend/Dockerfile.vars
```diff
@@ -1,31 +0,0 @@
...
-
-RUN npm run build
-
-# DEPLOY CLIENT
-FROM nginx:latest as production-stage
-WORKDIR /app
-COPY --from=build-stage /app/packages/client/build /app
-COPY ./deploy/nginx.conf /etc/nginx/nginx.conf
-EXPOSE 80
-RUN ls -la /app
```
germanurrus, line 31: @fersan1985 I don't see a reason why not deleting this....is it ok?

jbluks: Yea - I was looking at this and it seemed outdated and missing things from the other scripts. I deleted to avoid confusion. However, @fersan1985 can confirm if this file isn't used anywhere.

PR #475, opened 2022-09-15
CAS-504 + CAS-505: Add Portland Trailblazers and Sacramento Kings custom strategies

jbluks:
Adds Contracts for NBA TopShots Portland Trailblazers and Sacramento Kings.


rvmelkonian (approved):
good to go, note to iterate on the approach here by adding in string replace function for team names.

PR #483, opened 2022-09-15
Proposal Threshold cannot be < 1

tuffon:


jbluks (approved):

backend/main/server/controllers.go
```diff
@@ -391,6 +397,12 @@ func (a *App) updateCommunity(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
+	if payload.Proposal_threshold != nil && payload.Only_authors_to_submit != nil {
+		err = validateProposalThreshold(*payload.Proposal_threshold, *payload.Only_authors_to_submit)
+		if err != nil {
+			respondWithError(w, http.StatusBadRequest, err.Error())
+		}
+	}
```
rvmelkonian, line 405: and same with this block

backend/main/community_test.go
```diff
@@ -66,13 +66,26 @@ func TestCreateCommunity(t *testing.T) {
...
+	checkResponseCode(t, http.StatusBadRequest, response.Code)
+}
+
+func TestCreateCommunityFailThreshold(t *testing.T) {
+	// Prep
+	clearTable("communities")
+	clearTable("community_users")
+
+	// Create Community
+	communityStruct := otu.GenerateFailThresholdCommunityStruct("account")
```
rvmelkonian, line 88: thanks for the tests!

rvmelkonian (commented):
looks good! Just one small suggestion.

backend/main/server/controllers.go
```diff
@@ -359,6 +359,12 @@ func (a *App) createCommunity(w http.ResponseWriter, r *http.Request) {
...
```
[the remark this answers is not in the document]
tuffon: So I wanted to move as much logic into the helper function as I could by passing the entire payload into the function and then validating but because the shape of the payload for create and update are different it didn't quite work out the way I wanted it to, I found that validating the nil pointer ref and then passing the values was the cleanest way to do so without writing separate functions like `validateCreateCommunityProposalThreshold` and `validateUpdateCommunityProposalTheshold` which I guess an argument could be made for that approach, however comparatively I prefer this way better.  I'm open to a cleaner pattern here as well, I have been looking around for cleaner patterns around nil pointer checking in go as I am still kinda relatively new to the language so i'm open to any suggestions!

backend/main/server/controllers.go
tuffon: resolving duplicate

PR #484, opened 2022-09-16
rich/err-fix update error code cas -520 

rvmelkonian:
•updates error code with `403` and more descriptive message if the user does not have sufficient balance

### DIFF
--- a/backend/main/server/controllers.go
+++ b/backend/main/server/controllers.go
@@ -135,7 +135,7 @@ func (a *App) createVoteForProposal(w http.ResponseWriter, r *http.Request) {
 
 	vote, err := helpers.createVote(r, proposal)
 	if err != nil {
-		respondWithError(w, http.StatusInternalServerError, err.Error())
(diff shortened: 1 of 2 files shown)


backend/main/server/controllers.go
```diff
@@ -135,7 +135,7 @@ func (a *App) createVoteForProposal(w http.ResponseWriter, r *http.Request) {
 
 	vote, err := helpers.createVote(r, proposal)
 	if err != nil {
-		respondWithError(w, http.StatusInternalServerError, err.Error())
+		respondWithError(w, http.StatusForbidden, err.Error())
```
mannynotfound, line 138: if `helpers.createVote` hits the lines below then we will return forbidden when really its just an invalid payload

https://github.com/DapperCollectives/CAST/blob/ceddc9a86bf85bddfe934dcd0687379c7e39af03/backend/main/server/helpers.go#L288-L292

jbluks:
This PR is handled in https://github.com/DapperCollectives/CAST/pull/495

PR #485, opened 2022-09-16
Rich/cas-530 Unable to cancel active custom-script proposals

rvmelkonian:
• removed hexify logic on BE as timestamp was already being hexified on FE.


rvmelkonian:
@germanurrus do you mind giving this a test?

germanurrus (approved):
Tested locally and works as expected.

sarahy-Dapper (approved):

PR #495, opened 2022-09-16
CAS-520: Fix error code return when creating a proposal

jbluks:
- Updates helper to handle different error conditions and return the correct status


mannynotfound (approved):
LGTM

germanurrus (approved):
LGTM

PR #498, opened 2022-09-16
fix payload for removing users 

germanurrus:


frontend/packages/client/src/hooks/useCommunityUsersMutation.js
```diff
@@ -80,11 +80,11 @@ export default function useCommunityUsersMutation({ communityId } = {}) {
   const addCommunityUsers = useCallback(
     async ({ userType, addrs, body }) => {
       return new Promise((resolve, reject) => {
-        addrs.forEach((addrToRemove) => {
```
germanurrus, line 83: change in the name only

jbluks: Nice - saw that when I was looking at this issue.

jbluks (approved):

PR #501, opened 2022-09-19
CAS-545: Fix flow address validation for custom-script

jbluks:


germanurrus (approved):
Tested locally and it works as expected.

PR #503, opened 2022-09-19
CAS-547: Fix nil and undefined errors requesting votes for proposals

jbluks:
- Fix undefined error on the FE when there are no proposals yet in a community
- Fix nil error on the BE where the strategy was not being initialized when retrieving votes for proposal


rvmelkonian (approved):
missing the `InitStrategy` I see, yeah looks good.

PR #512, opened 2022-09-21
CAS-485 - Proposal view results & information

germanurrus:
- Update for proposal panel information

Screen Shot 2022-09-20 at 21 11 23
Screen Shot 2022-09-20 at 21 11 47


frontend/packages/client/src/App.sass
```diff
@@ -15,7 +15,7 @@ $family-monospace: "DM Sans"
 $body-color: black
 $body-background-color: white
 $primary: $yellow
-$spacing-values: ("0": 0, "1": 0.25rem, "2": 0.5rem, "3": 0.75rem, "4": 1rem, "5": 1.5rem, "6": 3rem, "7": 4.5rem,"8": 6rem, "9": 9rem, "auto": auto) !default
+$spacing-values: ("0": 0, "1": 0.25rem, "2": 0.5rem, "3": 0.75rem, "4": 1rem, "5": 1.5rem, "5_5": 2.5rem, "6": 3rem, "7": 4.5rem,"8": 6rem, "9": 9rem, "auto": auto) !default
```
germanurrus, line 19: This is not the most happy way of doing but it looks like there's a lot of components that require padding of 40px

frontend/packages/client/src/components/ProposalInformation/InfoBlock.js
```diff
@@ -0,0 +1,32 @@
...
+        }}
+      >
+        {content}
+        {component && cloneElement(component, { ref: containerRef })}
+      </div>
+    </div>
+  );
+};
+
+export default InfoBlock;
```
germanurrus, line 32: This is an existing component, just extracted this into a file with some teaks

frontend/packages/client/src/components/ProposalInformation/BlockieWithAddress.js
```diff
@@ -0,0 +1,43 @@
...
+            <span className="pr-2">{addr}</span>
+            <Svg name="LinkOut" width="12" height="12" />
+          </div>
+        </a>
+      </div>
+    </div>
+  );
+});
+
+export default BlockieWithAddress;
```
germanurrus, line 43: This is an existing component, just extracted this into a file with some teaks

frontend/packages/client/src/components/ProposalInformation/Information.js
```diff
@@ -0,0 +1,150 @@
...
+                <Svg name="LinkOut" width="12" height="12" />
+              </Tooltip>
+            </a>
+          }
+          isLastElement
+        />
+      )}
+    </div>
+  );
+}
```
germanurrus, line 150: this was part of the ProposalInformation component, but since that component has a lot of logic for sticking the panel and also loading data, this one was created to show data only

frontend/packages/client/src/components/ProposalInformation/ProposalInformation.js
```diff
@@ -0,0 +1,177 @@
...
+          maxWeight={maxWeight}
+          minBalance={minBalance}
+          openStrategyModal={openStrategyModal}
+        />
+      </div>
+    </div>
+  );
+};
+
+export default ProposalInformation;
```
germanurrus, line 177: This was an existing component, moved to another location and extracted components from it

frontend/packages/client/src/components/ProposalInformation/ProposalInformation.js
```diff
@@ -0,0 +1,177 @@
...
+
+  const {
+    name: tokenName,
+    maxWeight: maxWeightFromContract,
+    minBancale: minBalancefromContract,
+  } = contract;
+
+  // get maxWeight and minBalance from proposal or default to community settings
+  const maxWeight = proposalMaxWeight ?? maxWeightFromContract;
+  const minBalance = proposalMinBalance ?? minBalancefromContract;
```
germanurrus, line 70: making this to take max and min from proposal first and later from community contract

frontend/packages/client/src/components/ProposalInformation/Information.js
```diff
@@ -0,0 +1,150 @@
...
+              {strategyName}
+            </div>
+          </div>
+        }
+      />
+      <InfoBlock
+        title={'Token required'}
+        component={
+          <a
+            href={'https://flowscan.org/'}
```
germanurrus, line 81: need to discuss with Product where this link will point

frontend/packages/client/src/components/ProposalInformation/Information.js
```diff
@@ -0,0 +1,150 @@
...
+      />
+
+      <InfoBlock
+        title={'Proposed by'}
+        component={
+          <BlockieWithAddress
+            creatorAddr={creatorAddr}
+            isCoreCreator={isCoreCreator}
+          />
+        }
```
germanurrus, line 65: need to discuss with Product where this link will point

frontend/packages/client/src/components/ProposalInformation/ProposalInformation.js
```diff
@@ -0,0 +1,177 @@
...
+
+  // this effect watches for user scroll to make info panel fixed to navbar
+  useEffect(() => {
```
jbluks, line 73: Can this be converted into a hook that we can reuse? If this is specific to this component, can we add to a hooks file and put all this logic there and just call the hook here?

frontend/packages/client/src/components/ProposalInformation/ProposalInformation.js
```diff
@@ -0,0 +1,177 @@
...
+        topRef.current.pointStatic > winScroll
+      ) {
+        topRef.current.pointStatic = null;
+        setFixedStyle(null);
+      }
+    }
+  }, [scroll, fixedStyle, windowHeight, navbarHeight]);
+
+  // this effect watches for window width resizes to change info panel width
+  useEffect(() => {
```
jbluks, line 115: Same here - can we extract this into a hook that we call here?

frontend/packages/client/src/components/ProposalInformation/ProposalInformation.js
```diff
@@ -0,0 +1,177 @@
...
+      }
+    }
+  }, [fixedStyle, windowWidth, navbarHeight]);
+
+  function handleScroll() {
+    setScroll(document.body.scrollTop || document.documentElement.scrollTop);
+  }
+
+  // this effect watches for window scrolling
+  useEffect(() => {
```
jbluks, line 135: Same here.

frontend/packages/client/src/components/ProposalInformation/ResultsPanel.js
```diff
@@ -0,0 +1,175 @@
+import { useMemo } from 'react';
+import { Svg } from '@cast/shared-components';
+import { Pill } from 'components';
+import { FilterValues } from 'const';
+import { parseDateFromServer } from 'utils';
+
+const Results = ({ voteResults, computedStatus }) => {
```
jbluks, line 7: This looks like we have two components in here - is it possible to separate these into separate component files?

jbluks (changes requested):

frontend/packages/client/src/components/ProposalInformation/ProposalInformation.js
germanurrus: This is very specific for this info panel component, it makes the panel to be sticked to the nav bar when you scroll down, I think could be managed as a tech debt, I only copy the lines from the previous component before extracting other components that were in the same file.

frontend/packages/client/src/components/ProposalInformation/ProposalInformation.js
germanurrus: same for this we could create a tech-debt issue to address this

frontend/packages/client/src/components/ProposalInformation/ResultsPanel.js
germanurrus: yes, makes sense. I just created a new component

germanurrus:
@jbluks PR updated

jbluks (approved):

PR #513, opened 2022-09-21
CAS-528 - Update error modal

germanurrus:
Removed Error component and replace it with ErrorModal 
Updated calls for `createElement(Error, { ...})` with  `createElement(ErrorModal, { ...})`
Updated calls for `notifyError()`

Screen Shot 2022-09-21 at 12 32 26

Screen Shot 2022-09-21 at 12 33 18

Screen Shot 2022-09-22 at 17 54 30

Screen Shot 2022-09-22 at 19 39 06

PR #523, opened 2022-09-22
cas-527 updated error messages and handling

rvmelkonian:
•updates error messages by creating controller level constants as discussed with German, the heading and body message are separated by double colons for easy parsing in the client


jbluks:
I am wondering if we can use this ticket to implement some best practices in our error handling. Some suggestions I have:

1. Let's make our error codes follow a standard like `ERR_1000`, `ERR_1001`, etc. This keeps all error codes the same format/length and are easily readable.

2. Can we create an ErrorCode struct that we return as JSON with the following format:

```
type ErrorResponse struct {
  code        string
  message string
  details     string
}
```

The message is what can be used as the title of the modal, and the details used as the body of the modal. Can we just use JSON instead of a custom format with delimiters?

3. Can we define friendly constants for the errors in our code? eg. `ErrIncompleteRequest`, `ErrCreateCommunity`, `ErrFetchBalance`, etc. I am concerned that using the error codes as variable names makes the code less readable.

4. If we don't want to go with the ErrorCode struct, can we declare the above variables as `errors.New()` or perhaps better `fmt.Errorf()` so we can just pass them straight to respondWithError without having to call `errors.New()` every time? I would like to see something like `respondWithError(w, Status, ErrIncompleteRequest.Error())` 

5. One thing I saw in a few blog posts is recommending the use of `fmt.Errorf()` over `errors.New()` (e.g. [this post](https://levelup.gitconnected.com/go-error-best-practice-f0864c5c2385)). Could we adopt that as a standard for our code?

rvmelkonian:
thanks for this feedback! The format of `ERR001::header::body` was only decided upon as @germanurrus asked for me to implement it this way so that he can easily parse it on the front end. 

1. Yes happy to use this format for the constants.

2. I'm up for either JSON or using `fmt.Errorf()`, perhaps we to use `fmt.Errorf()` as I think the front end code to parse `ERR001::header::body` this format have already been written? Unless we would all prefer JSON. I think I actually don't mind either one. 

3. Yes.

4. Yes, that's a good idea.

5. - Yes

germanurrus:
@rvmelkonian @jbluks I don't have any hard feelings with using a JSON format. I'm using a function to parse the string, but if that becomes a JSON that can be easily updated: https://github.com/DapperCollectives/CAST/pull/513/files#diff-96d40116d41a7df3efad28fc0649d0596486ab914aa726596521fbec0a19ce5dR16

backend/main/server/controllers.go
```diff
@@ -14,6 +13,52 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+type errorResponse struct {
+	statusCode int
+	modalCode  string
```
jbluks, line 18: @rvmelkonian can we update this to be `errorCode` or something similar? I think calling it `modalCode` doesn't make sense since we don't care how the front-end consumes errors. Some errors could be consumed inline for example.

jbluks (changes requested):

jbluks (approved):

backend/main/server/controllers.go
jbluks: Updated.

jbluks (approved):

jbluks:
@rvmelkonian This should be targetting release-v1.5.0

PR #529, opened 2022-09-26
added pagination for search endpoint

rvmelkonian:
•adds pagination for search endpoint


jbluks (approved):

jbluks (changes requested):
Please change to target release-v1.5.0.

jbluks (approved):

PR #532, opened 2022-09-26
CAS-567 Fix on user rejecting signing

germanurrus:


frontend/packages/client/src/utils.js
```diff
@@ -75,8 +76,10 @@ export const getCompositeSigs = (sigArr) => {
   if (sigArr[0]?.signature?.signature) {
     return [sigArr[0].signature];
   }
-
-  if (typeof sigArr === 'string' && sigArr.includes('Declined:')) {
+  if (
+    isError(sigArr) ||
+    (typeof sigArr === 'string' && sigArr.includes('Declined'))
```
germanurrus, line 81: looks like fcl is returning an error object when user is connected with blocto and rejects signing for lilico they are doing the same thing as before

PR #541, opened 2022-09-26
CAS-550: Ranked Choice Voting

jbluks:


backend/main/strategies/tally_methods.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/DapperCollectives/CAST/backend/main/models"
+	"golang.org/x/exp/slices"
+)
+
+type TallyStruct struct {
+	choice string
+	votes int
+}
+
+// TODO: What if proposal is not closed yet? Do we tally for a winner, or just display the total first choice votes for each choice?
```
dbslone, line 15: Check with Jose on this my guess is we on the FE we wouldn't show the results until after the proposal closes

dbslone: @jbluks how do we determine if the user selected the winning vote for the leaderboard?

jbluks: @dbslone I think is should just work, since we are checking the first place vote anyways regardless of single choice/ranked choice. And only a first place vote for a winning choice should count as a winning achievement bonus.

jbluks:
This PR has been merged into: https://github.com/DapperCollectives/CAST/pull/571

PR #544, opened 2022-09-27
added validateStrategyName and error handle

rvmelkonian:
•adds `validateStrategyName` function
•exports StrategyMap 
•adds in new error code


rvmelkonian:
conflicts on here are because I accidentally made this branch from main (out of habit), will fix these!

PR #545, opened 2022-09-28
cas-577 catch strategy name errors. update tests

rvmelkonian:
•Adds new `errorResponse : errStrategyNotFound`
•Changes return type of `createProposal` for better error handling.

### DIFF
--- a/backend/main/server/controllers.go
+++ b/backend/main/server/controllers.go
@@ -78,6 +78,8 @@ var (
 		message:    "Strategy Not Found",
 		details:    "The strategy name you are trying to use no longer exists.",
 	}
+
+	nilErr = errorResponse{}
 )
 
 func (a *App) health(w http.ResponseWriter, r *http.Request) {
@@ -308,7 +310,7 @@ func (a *App) createProposal(w http.ResponseWriter, r *http.Request) {
 	}
 
 	proposal, errResponse := helpers.createProposal(p)
-	if err != nil {
+	if errResponse != nilErr {
 		log.Error().Err(err).Msg("Error creating proposal")
 		respondWithError(w, errResponse)
 		return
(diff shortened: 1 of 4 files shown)


backend/main/server/controllers.go
```diff
@@ -300,10 +307,10 @@ func (a *App) createProposal(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	proposal, err := helpers.createProposal(p)
+	proposal, errResponse := helpers.createProposal(p)
 	if err != nil {
```
jbluks, line 311: This won't catch because it is checking for err instead of errResponse

backend/tests/proposal_test.go
```diff
@@ -37,12 +37,14 @@ func TestGetProposal(t *testing.T) {
 	t.Run("Should throw an error if proposal with ID doesnt exist", func(t *testing.T) {
 
 		response := otu.GetProposalByIdAPI(communityId, 420)
+		fmt.Printf("Response: %s", response.Body)
```
jbluks, line 40: Remove println

backend/tests/proposal_test.go
```diff
@@ -37,12 +37,14 @@ func TestGetProposal(t *testing.T) {
...
 
 		CheckResponseCode(t, http.StatusForbidden, response.Code)
 
-		var m map[string]string
-		json.Unmarshal(response.Body.Bytes(), &m)
-		assert.Equal(t, "ERR_1001", m["errorCode"])
+		var e errorResponse
+		json.Unmarshal(response.Body.Bytes(), &e)
+		fmt.Printf("Error: %v", e)
```
jbluks, line 46: If this print is not necessary, remove

backend/tests/proposal_test.go
```diff
@@ -160,6 +162,8 @@ func TestUpdateProposal(t *testing.T) {
 		var p models.Proposal
 		json.Unmarshal(response.Body.Bytes(), &p)
 
+		fmt.Printf("Proposal ID: %d", p.ID)
```
jbluks, line 165: Remove

jbluks (changes requested):
Some small clean ups, and one error

backend/main/server/controllers.go
rvmelkonian: got around this by creating `nilErr` of type `errorResponse{}`, then we can do the conditional check against that.

jbluks (approved):

PR #547, opened 2022-09-28
CAS-580: Vote/Proposal Choices Refactor

jacksonConrad:
- Set ID on proposal Choices
- Store choice ID when creating vote on FE
- Use choice ID to calculate and display proposal results
- update affected tests

### DIFF
--- a/frontend/packages/client/src/components/ProposalInformation/Results.js
+++ b/frontend/packages/client/src/components/ProposalInformation/Results.js
@@ -1,7 +1,13 @@
 import { useMemo } from 'react';
 import { FilterValues } from 'const';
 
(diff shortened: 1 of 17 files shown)


frontend/packages/client/src/components/ProposalInformation/Results.js
```diff
@@ -1,7 +1,13 @@
 import { useMemo } from 'react';
 import { FilterValues } from 'const';
 
-export default function Results({ voteResults, computedStatus } = {}) {
+export default function Results({
+  proposalChoices,
+  voteResults,
+  computedStatus,
+} = {}) {
+  console.log('results.js');
```
jbluks, line 9: remove console.log

PR #558, opened 2022-09-29
Implement Sentry.io

dbslone:


frontend/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "dependencies": {},
   "resolutions": {
-    "babel-loader": "8.1.0"
+    "babel-loader": "8.1.0",
+    "react-error-overlay": "6.0.9"
```
dbslone, line 18: This fixes the issue with react-error-overlay emitting `process is not defined` error. https://stackoverflow.com/questions/70368760/react-uncaught-referenceerror-process-is-not-defined

germanurrus (approved):
LGTM! 🚀

PR #569, opened 2022-10-04
CAS-593: Fix modal error message for insufficient balance

jbluks:


backend/main/server/controllers.go
```diff
@@ -217,8 +245,8 @@ func (a *App) createVoteForProposal(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	vote, err := helpers.createVote(r, proposal)
-	if err != nil {
+	vote, errResponse := helpers.createVote(r, proposal)
+	if errResponse != nilErr {
 		log.Error().Err(err).Msg("Error creating vote.")
 		respondWithError(w, errInsufficientBalance)
```
germanurrus, line 251: I think here the string is not getting the params to interpolate: 
`Details: `In order to vote on this proposal you must have a minimum 
							balance of %d %s tokens in your wallet.`,`

jbluks: Found the issue - the controller wasn't returning the errResponse, but the constant instead.

germanurrus (approved):

PR #574, opened 2022-10-05
CAS-541: Search Endpoint

jbluks:
- Fix the endpoint to be `/search?text=search+text` to allow for spaces. If `text` is blank, the default search for featured communities will be performed
- Bugs in the filter counts also fixed.

### DIFF
--- a/backend/main/server/helpers.go
+++ b/backend/main/server/helpers.go
@@ -470,7 +470,7 @@ func (h *Helpers) fetchCommunity(id int) (models.Community, error) {
 }
 
 func (h *Helpers) searchCommunities(
-	query string,
+	searchText string,
 	filters string,
 	pageParams shared.PageParams,
 ) (
@@ -478,7 +478,7 @@ func (h *Helpers) searchCommunities(
 	map[string]int,
 	error,
 ) {
-	if query == "defaultFeatured" {
+	if searchText == "" {
 		isSearch := true
 
 		results, _, err := models.GetDefaultCommunities(
@@ -496,16 +496,17 @@ func (h *Helpers) searchCommunities(
 
 		return results, categories, nil
 	} else {
+		fmt.Println(filters)
 		filtersSlice := strings.Split(filters, ",")
 		results, err := models.SearchForCommunity(
 			h.A.DB,
-			query,
+			searchText,
 			filtersSlice,
 		)
 		if err != nil {
 			return []*models.Community{}, nil, err
 		}
-		
+
 		categories := h.categoryCountToMap(results)
 		return results, categories, nil
 	}
--- a/backend/main/server/helpers.go
+++ b/backend/main/server/helpers.go
@@ -486,6 +486,7 @@ func (h *Helpers) searchCommunities(
 			pageParams,
 			isSearch,
 		)
+
 		if err != nil {
 			log.Error().Err(err)
 			return nil, nil, err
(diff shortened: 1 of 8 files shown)


frontend/packages/client/src/pages/BrowseCommunities.js
```diff
@@ -41,16 +41,12 @@ export default function BrowseCommunities() {
```
germanurrus, line 41: we need this one to be updated (Line 29):
```
const { search, filters: filtersUrl } = useQueryParams({
    search: 'search',
    filters: 'filters',
  });
  const [searchText, setSearchText] = useState(search ? search : '');
  ```
  to be 
```
const { text, filters: filtersUrl } = useQueryParams({
    text: 'text',
    filters: 'filters',
  });
  const [searchText, setSearchText] = useState(text ? text : '');
  
  ```
  This will enable copy paste search url and you'll get the same search results and also navigating back from a community

backend/main/server/helpers.go
```diff
@@ -469,13 +469,67 @@ func (h *Helpers) fetchCommunity(id int) (models.Community, error) {
...
+	return categoryCount
+}
+
+func (h *Helpers) getCategoryCount(results []*models.Community) []interface{} {
+	var resultsWithCount []interface{}
+	for _, community := range results {
+		resultsWithCount = append(resultsWithCount, community)
+	}
+
+	return resultsWithCount
```
germanurrus, line 532: It looks like the count url param is not taken into account on the search result, I've created a set for testing and results array has more elements than the count number on the first pull from the backend (FE sends 10 and backend returns all results):

Screen Shot 2022-10-06 at 11 21 34

The other thing I noticed is that results are not filtered with the text search after the space, in this case "before 2" should return only the community that has "before 2" but it returns the same as if I would enter "before" (The name of the community makes no sense :( ). May it has to do with the configuration of the search plug-in?

germanurrus (changes requested):
Looking good! Made some comments

backend/main/server/helpers.go
jbluks: Fixed the search results count/start. Was missing from the search text query. The results are inclusive of any of the words, so because the search contains "before" it will return any community with that word, even though the search contains another search string. The search does not "narrow" with new terms. I think this makes sense for a browser page.

germanurrus:
there's one last glitch while filtering, if the frontend sends the filter results should be narrowed but the filter should return the same amount originally created for the text search: 

https://user-images.githubusercontent.com/7526740/194403104-efba0a7a-4392-4d4c-955a-536efd1346d7.mov

frontend/packages/client/src/hooks/useCommunitySearch.js
```diff
@@ -54,12 +54,13 @@ export default function useCommunitySearch({
...
           ...(filters ? { filters: filters } : undefined),
           start,
           count,
         }).toString();
 
-        const url = `${process.env.REACT_APP_BACK_END_SERVER_API}/communities/search/${searchText}?${queryParams}`;
+        const url = `${process.env.REACT_APP_BACK_END_SERVER_API}/communities/search?${queryParams}`;
 
         const response = await fetch(url);
         return checkResponse(response);
```
germanurrus, line 66: on line 70: 
```
 getNextPageParam: (lastPage) => {
          // custom adjustment
          const { next, start, count, totalRecords } = lastPage?.results ?? {};
          return [start + count, count, totalRecords, next];
        },

germanurrus (approved):
Tested locally and works as expected!

PR #595, opened 2022-10-07
Linxiaoli/vsl 253: rebase and create useNotification hook

linxiao-li:
contains the change of latest main
adds useNotificationService hook


frontend/packages/client/src/hooks/useNotificationService.js
```diff
@@ -0,0 +1,106 @@
...
+  const getUserSettings = async () => {
+    try {
+      //here we call api
+      setNotificationSettings();
+    } catch {
+      throw new Error('cannot get user settings');
+    }
+  };
+
+  const unsubscribeCommunityNotifications = async (communityId) => {
```
amyliu6, line 43: maybe rename to `unsubscribeCommunity` to have consistent function naming with `subscribeCommunity`

germanurrus (commented):
Is this being ported to Vessel to CAST? or is this a brand new hook implemented on CAST? The reason I'm asking is because I was looking at the `useNotificationService`  hook and this would look more like a context providing the state and functions, or maybe this hook will be called from a context, is there place we can see the implementation plan? thanks!

linxiao-li:
@germanurrus My original thought is to have the components calling the hook methods and manage the api subscription using useEffect, thanks for getting me to rethink about using context! Upon reading the leanplum documentation, it recommends start the session upon loading the whole application, context for sure works better in this case. 
I've created a NotificationServiceContext and have the useNotificationService hook use it. Example implementations can be found on WalletConnect component and Settings Page. Let me know if make more sense!

frontend/packages/client/src/App.sass
```diff
@@ -303,15 +305,16 @@ hr
...
 	.section
 		padding-right: 1rem
 		padding-left: 1rem
 	.navbar-burger
 		display: block
 		width: 1.25rem
-	.navbar-brand
-		flex-direction: row-reverse
+		span
+			height: 2px !important
```
mannynotfound, line 317: I think this PR needs to be rebased off main because this seems like an unrelated change that's already in main branch

amyliu6: maybe let's try to decouple `rebase` and feature work. Create separate PR to merge new changes from 'v1.6' into `notification` and then our local branch only pull from `notification` so PR should only contain the new feature changes. @mannynotfound  @linxiao-li  what do you think?

amyliu6: opened a PR here to merge v1.6 to notifications: https://github.com/DapperCollectives/CAST/pull/603. above change is one of the diff carried over.

linxiao-li:
@amyliu6 @mannynotfound @germanurrus I pulled from the latest notifications, all the changes should be clean now. Please give me a thumbs up if things look ok, then I'll go ahead and merge it

frontend/packages/client/src/components/WalletConnect.js
```diff
@@ -62,6 +67,11 @@ const SignInOutButton = ({
     return () => clearTimeout(timeout);
   }, [addressCopied]);
 
+  useEffect(() => {
+    setUserID(addr);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [addr]);
```
mannynotfound, line 73: this seems a little odd - why not just pass `addr` to `useNotificationService()`? I dont think we need to fetch arbitrary wallet addresses yet, just the current connected wallet.

mannynotfound (commented):
looks pretty good! just some small suggestions and a question

frontend/packages/client/src/components/WalletConnect.js
linxiao-li: Good question, the reason is exactly that: we only care about current connected wallet, when instantiated, `useNotificationService` reads from the context, we don't need to pass the address as part of the configuration. If it helps, consider the following example:

- We want to build a hook to produce apple, 
```
const useProduceApple = () =>{
    const seed = () =>{
        console.log("seeding apple")
    }
    const grow = ()=>{
      console.log("growing apple")
    }
    const harvest = ()=>{
        console.log("harvesting apple")
    }
    return {
        seed,
        grow,
        harvest
    }
}
```
- Now we want to grow banana too (the structure you are referring to)
```
const useProduceFruit = (fruitName) =>{
    const seed = () =>{
        console.log("seeding" + fruitName)
    }
    const grow = ()=>{
      console.log("growing" + fruitName)
    }
    const harvest = ()=>{
        console.log("harvesting" + fruitName)
    }
    return {
        seed,
        grow,
        harvest
    }
}
```

frontend/packages/client/src/hooks/useNotificationService.js
```diff
@@ -0,0 +1,105 @@
...
+    notificationSettings,
+    setUserID,
+    setUserEmail,
+    getUserSettings,
+    unsubscribeCommunity,
+    subscribeCommunity,
+    unsubscribeFromEmailNotifications,
+    resubscribeFromEmailNotifications,
+  };
+}
```
germanurrus, line 105: All these functions could be moved to the context provider and then returned  here https://github.com/DapperCollectives/CAST/pull/595/files#diff-992c4fc25965e4d49bc051c773bb74af39153b25c9936d99d25efcd86fc33c1dR24 as an object. 
This way all the logic is encapsulated in the provider, then on the component that is consuming you just need to call `useNotificationServiceContext()` and you get all the functions you need. 
An example could be the `contexts/ErrorHander` all functions are wrapped with useCallback and then the object passed to `ErrorHandlerContext.Provider` is wrapped with a useMemo. This is to prevent extra renders. Hope this is helps :)

frontend/packages/client/src/components/WalletConnect.js
germanurrus: I think we don't need to set the `addr` this way, the SignInOutButton component is using a HOC, at the end of the file is wrapped by Web3Consumer. 

`export default Web3Consumer(CurrentUser);`
so addr is passed from the Web3Consumer to the SignInOutButton and inside we are calling the useNotificationService to set the userId. 

There's this other way that I thinks is more direct: 
Taking a look at the structure: 
```
 <Web3Provider network={process.env.REACT_APP_FLOW_ENV}>
          <NotificationModalProvider>
            <NotificationServiceProvider>
              <Router>
                <ErrorHandler>
                  <AppPages />
                </ErrorHandler>
              </Router>
            </NotificationServiceProvider>
          </NotificationModalProvider>
        </Web3Provider>
  ```
The Web3Provider is wrapping all components, so we can do inside the `NotificationServiceProvider`: 
```
import { useWebContext } from 'contexts/Web3';
.
.
.
const [notificationSettings, setNotificationSettings] = useState({
    walletId: '',
    email: '',
    communitySubscription: [],
    isUnsubscribedFromCommunityUpdates: false,
  });

 const { user: { addr }  } = useWebContext();  

useEffect(()=> {
  if(addr) {
   setNotificationSettings(prev => ({
      ...prev,
      walletId: addr,
    }))
  }
  if(!addr && notificationSettings.walletId !== ''){
    // cleaning state bc user got disconnected
      setNotificationSettings({
          walletId: '',
          email: '',
          communitySubscription: [],
         isUnsubscribedFromCommunityUpdates: false,  
        } <== all of this could be a static object since it's used in the initialization  too 
     )
   }
}, [addr])

```
Then we don't need to make SignInOutButton component aware of the existence of this logic of the `NotificationServiceProvider`. We are coupling directly NotificationServiceProvider with the Web3Context provider directly and not through the SignInOutButton. 

When a user connects/disconnects it causes the re-render of the components that are consuming through the context so this will make the useEffect to run and set/un-set the state in the `NotificationServiceProvider`
Let me know if you have any questions :)

frontend/packages/client/src/hooks/useNotificationService.js
linxiao-li: All functions moved to context!
I have a different onion on using useCallback and useMemo though, so I profiled the http://localhost:3000/#/settings page with/without the useCallback and useMemo

- why settings page:
1. The Settings component calls useNotificationServiceContext hook only and renders a single string, the performance of this hook is isolated for comparison.
2. This page contains WalletConnect component which calls useNotificationServiceContext first, so theoretically by the time Settings component calls useNotificationServiceContext hook, all the methods should be cached and speed up the page load

- result: loading time is about the same with or without the optimization
With optimization:
Screen Shot 2022-10-13 at 10 20 13 AM
Without optimization:
Screen Shot 2022-10-13 at 10 18 31 AM

Here are some blog posts I found that explains when and where to use usecallback and usememo for performance:
https://dev.to/milu_franz/demystifying-react-hooks-usecallback-and-usememo-1a8j
https://kentcdodds.com/blog/usememo-and-usecallback

frontend/packages/client/src/components/WalletConnect.js
linxiao-li: Yes this approach is way better! Updated!

frontend/packages/client/src/hooks/useNotificationService.js
germanurrus: yeah I think in this case is fine not using them, because there's only one state that is holding all data, any update will trigger a render since we are saving all in one place, but in case there's a state is not passed down to children, updating state will cause an extra not needed rended.

PR #605, opened 2022-10-11
CAS-590: Fix Proposal information for custom scripts

jbluks:
- For custom scripts the `Required Token` field is not relevant. Instead, display the name/description of the custom script.

Screen Shot 2022-10-11 at 3 51 32 PM

PR #607, opened 2022-10-12
CAS-629: Token Required value does not display correctly for mobile

germanurrus:


germanurrus:
fixed by https://github.com/DapperCollectives/CAST/pull/605

PR #628, opened 2022-10-14
CAS-640 filter count bug fix

0xmovses:
• adds new method for generating correct Count SQL for totalRecords when filters are applied

### DIFF
--- a/backend/main/models/community.go
+++ b/backend/main/models/community.go
@@ -514,6 +514,7 @@ func generateDefaultFilterCountSql(filters []string) (string, error) {
 		var sql string = `
 				SELECT COUNT(*) FROM communities
         WHERE category IS NOT NULL
+				AND is_featured = true
 				AND category IN (`
 		for i, filter := range filters {
 			if i == len(filters)-1 {


germanurrus:
This are the combinations: 
- Featured (default search): this works :white_check_mark:

- Featured search with filters: Doesn't work :x: 
next never becomes -1 when there are no more results

https://user-images.githubusercontent.com/7526740/195849854-c829ef48-0901-4b92-828d-04066ae06cff.mov


- Text Search with no filters:  Doesn't work :x:
Only returns first page but there are more results in the db

https://user-images.githubusercontent.com/7526740/195849749-59f20498-616d-4c56-acf0-d46bcd3d6aed.mov


- Text Search with Filters:  this works :white_check_mark:

germanurrus (changes requested):
Left a comment about the text search conditions that need some fixes

germanurrus:
> * Featured search with filters

Both are still not working on my side

germanurrus (approved):
All case tests working now!

PR #632, opened 2022-10-14
VSL-230 - setup subscribe/manage notifications modal

mannynotfound:
adds the following components for signing up / managing notifications settings:

- `components/Notifications/NotificationsModal`
- `components/Notifications/NotificationsManage`
- `components/Notifications/NotificationsSignUp`
- `components/Notifications/NotificationsError`
- `components/Notifications/FormConfig`

This PR does not include an entry point so you'll have to add a button somewhere to launch the modal BUT these are what they should look like:

image
image
image
image


frontend/packages/client/src/App.js
```diff
@@ -30,15 +30,15 @@ function App() {
...
-              <Router>
+              <NotificationModalProvider>
                 <ErrorHandler>
                   <AppPages />
                 </ErrorHandler>
-              </Router>
+              </NotificationModalProvider>
             </NotificationServiceProvider>
-          </NotificationModalProvider>
+          </Router>
```
mannynotfound, line 41: move `<Router>` to wrap the modal provider so that the modal can link out to other parts of the app

frontend/packages/client/src/App.sass
```diff
@@ -582,7 +583,16 @@ span[data-tooltip]
...
+	opacity: 0
+
+.fade-out-half
+	-webkit-transition: opacity 0.3s ease-out
+	-moz-transition: opacity 0.3s ease-out
+	-ms-transition: opacity 0.3s ease-out
+	-o-transition: opacity 0.3s ease-out
+	opacity: 0.5
+	&.is-disabled
+		pointer-events: none
```
mannynotfound, line 595: there wasn't a design for the "loading" state of the email signup process - eg after you click signup but before the request has finished, so I just reduce the opacity of the entire modal and disable click events to indicate a loading state

frontend/packages/client/src/components/Input.js
```diff
@@ -5,11 +5,12 @@ export default function Input({
   value,
   disabled,
   onChange = () => {},
+  className = '',
 } = {}) {
   return (
     <input
       type="text"
-      className="rounded-sm border-light p-3 column is-full is-full-mobile mt-4"
+      className={`rounded-sm border-light p-3 column is-full is-full-mobile ${className}`}
```
mannynotfound, line 13: layout props like margin should be passed in by parent and not hardcoded onto component

frontend/packages/client/src/contexts/NotificationService.js
```diff
@@ -88,13 +87,15 @@ const NotificationServiceProvider = ({ children }) => {
...
         ...prevState,
-        communitySubscription: {
+        communitySubscription: [
           ...prevState.communitySubscription,
-          [communityId]: true,
-        },
+          communityId,
+        ],
       }));
+      return apiResponse;
```
mannynotfound, line 98: these were being turned into objects but they should be arrays so I re-wrote them to use arrays and added a fake api response for testing purposes

frontend/packages/client/src/components/Community/StrategySelectorForm.js
```diff
@@ -49,7 +49,6 @@ export default function StrategySelectorForm({
   );
 
   const addNewStrategy = (newStrategyInfo) => {
-    const { contract } = newStrategyInfo;
```
mannynotfound, line 52: this was an unrelated eslint error that was bothering me lol

frontend/packages/client/src/components/Notifications/NotificationsError.js
```diff
@@ -0,0 +1,31 @@
...
+        <button
+          className="button rounded-lg is-outlined px-3 flex-1 mr-2"
+          onClick={onClose}
+        >
+          <b>Close</b>
+        </button>
+      </div>
+    </section>
+  </div>
+);
```
germanurrus, line 29: Can we make all these notification components use the Modal component so the share the same styles from the same component, for example the error notification is using another icon in this PR and on designs we got: https://www.figma.com/file/bni5O6c7Yh1rCUwFRO8f9i/CAST%3A-Design-Library?node-id=328%3A3475

This is the block builder: https://github.com/DapperCollectives/CAST/blob/release-v1.5.0/frontend/packages/client/src/components/modals/Modal.js

mannynotfound: yep let me adjust

mannynotfound:
@germanurrus updated to use `Modal` components and moved all the new files to `components/modals/Notifications`

frontend/packages/client/src/components/modals/Error.js
```diff
@@ -17,7 +17,7 @@ export default function ErrorModal({
       ) : (
         <Modal.FooterContainer>
           {faqLink ? (
-            <Modal.FotterWithFAQ faqLink={faqLink} onClose={onClose} />
+            <Modal.FooterWithFAQ faqLink={faqLink} onClose={onClose} />
```
mannynotfound, line 20: typo correct

frontend/packages/client/src/components/modals/Modal.js
```diff
@@ -10,7 +10,7 @@ export default function Modal({ children }) {
 Modal.Header = ({ title, onClose }) => {
   return (
     <div className="c-modal-header spacing-header">
-      <p className="c-modal-title">{title}</p>
+      <p className="c-modal-title is-capitalized">{title}</p>
```
germanurrus, line 13: can we have this `is-capitalized` as a prop?, we have other paces using without allcaps

frontend/packages/client/src/components/modals/Notifications/SignUp.js
```diff
@@ -0,0 +1,107 @@
...
+    </Modal>
+  );
+};
+
+export default function NotificationsSignUp(props) {
+  const [errorMessage, setErrorMessage] = useState('');
+
+  if (errorMessage) {
+    return <Error message={errorMessage} onClose={props.onClose} />;
+  }
```
germanurrus, line 104: this will always render the Error component bc errorMessage will act as a true value

frontend/packages/client/src/components/modals/Notifications/Manage.js
```diff
@@ -0,0 +1,40 @@
...
+        <Link
+          to="/settings"
+          className="button rounded-lg is-primary px-3 flex-1 mr-2"
+          onClick={onClose}
+        >
+          <b>Settings</b>
+        </Link>
+      </div>
+    </section>
+  </div>
```
germanurrus, line 37: can you also update this one to use Modal as with SignUp?

germanurrus (changes requested):
Thanks for the updates! left some comments

frontend/packages/client/src/components/modals/Modal.js
mannynotfound: @germanurrus this isn't all caps, this is sentence case (Titles Like This) which I believe should be universal throughout the app. Where would we not want this style?

frontend/packages/client/src/components/modals/Notifications/SignUp.js
mannynotfound: from my testing this wasn't the case, `errorMessage` is only set when user submits and then gets cleared when they close the modal

frontend/packages/client/src/components/modals/Modal.js
germanurrus: you are right, this should use the class `is-capitalized`, thanks for the clarification

frontend/packages/client/src/components/modals/Notifications/Manage.js
mannynotfound: oops dont know how I missed that - updated now

germanurrus (approved):
LGTM!

germanurrus (approved):
LGTM!

PR #638, opened 2022-10-14
Order Search by Similarity Score

jbluks:
Adds the similarity score to the query so the search results can be ordered by the most relevant matching communities.

### DIFF
--- a/backend/main/models/community.go
+++ b/backend/main/models/community.go
@@ -12,6 +12,7 @@ import (
 	s "github.com/DapperCollectives/CAST/backend/main/shared"
 	"github.com/georgysavva/scany/pgxscan"
 	"github.com/jackc/pgx/v4"
+	"github.com/rs/zerolog/log"
 )
 
 type Community struct {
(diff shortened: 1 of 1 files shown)


germanurrus (approved):

backend/main/models/community.go
```diff
@@ -180,10 +181,10 @@ const UPDATE_COMMUNITY_SQL = `
 	WHERE id = $21
 `
 const SEARCH_COMMUNITIES_SQL = `
-	SELECT id, name, body, logo, category	
+	SELECT id, name, body, logo, category, SIMILARITY(name, $1) as score	
 	FROM communities 
 	WHERE SIMILARITY(name, $1) > 0.1
-		AND category IS NOT NULL
+		AND category IS NOT NULL ORDER BY score DESC
```
dbslone, line 187: I don't think you can add the ORDER BY here because `addFiltersToSql` appends to this string

backend/main/models/community.go
```diff
@@ -465,10 +466,15 @@ func SearchForCommunity(
...
 	var communities []*Community
+	// score is required for scanning, but can be ignored. Only used
+	// to order the search results by SQL.
+	var score float32
 	for rows.Next() {
 		var c Community
-		err := rows.Scan(&c.ID, &c.Name, &c.Body, &c.Logo, &c.Category)
+		err := rows.Scan(&c.ID, &c.Name, &c.Body, &c.Logo, &c.Category, &score)
 		if err != nil {
+			fmt.Println(err)
```
dbslone, line 476: do we need this?

backend/main/models/community.go
jbluks: Mmmmm right...

PR #644, opened 2022-10-17
CAS 639 & CAS 642 : additional test cases for pagination, search functions refactor

0xmovses:
This PR adds more test cases for pagination for search and cleans up the search models methods. 

• updates `GetSearchCommunitesAPI` with param `start *int`

The cases are: 
`should have correct total records for 'default featured'`
`next should be '-1' when there are no more records for 'default featured'`
`should have correct total record for search term with no filters`
`next should be '-1' when there are no more records for search term with no filters`
`should have correct total record for search term with filters`
`next should be '-1' when there are no more records for search term with filters`
`should have correct total records with search term and multiple filters`
`next should be '-1' when there are no more records for search term and multiple filters`

• Creates a new file `community_search.go` which contains all the search methods and constants. 
• Adds `processTotalRecords` which removes old repeating logic and creates reusable function
• Adds `runTotalRecordsQuery` for DRYness
• Removes `GetDefaultCommunities` adds instead `GetSearchDefaultCommunities` and `GetHomePageCommunities`;  better separation of concerns.
• removes `-cover` flat from makefile Test command as it was redundant.


germanurrus (commented):
Tested locally and works as expected

backend/Makefile
```diff
@@ -10,7 +10,7 @@ run:
 	./${BINARY_NAME}
 
 test:
-	go test -v -cover ./...
+	go test -v ./tests
```
jbluks, line 13: Not sure if this will work, since I have added unit tests in the strategies directory. `./tests` contains integration based tests, but we may have unit tests in the other directories. Unless we want to structure the unit tests differently.

jbluks (changes requested):

PR #650, opened 2022-10-17
VSL-242 - add NotificationsModal entry points to community + proposal pages

mannynotfound:
- adds `SubscribeCommunityButton` component which is similar to `JoinCommunityButton` as they appear next to each other on the community/proposal pages

image
image
image


frontend/packages/client/src/App.sass
```diff
@@ -153,6 +153,8 @@ code
 	flex: 1 0 0
 .flex-2
 	flex: 2 0 0
+.has-text-yellow
+	color: $yellow
 .has-background-orange
```
mannynotfound, line 158: utility class so we dont have to hardcode the yellow value in `BellFilled`

frontend/packages/client/src/components/Community/JoinCommunityButton.js
```diff
@@ -112,7 +112,7 @@ export default function JoinCommunityButton({
         onClick={isMember ? leaveCommunity : joinCommunity}
       >
         <Svg name="Eye" />
-        <Svg name="HideEye" />
+        <Svg name="EyeHidden" />
```
mannynotfound, line 115: changed the convention to `Icon<State>` so files are easier to read. This matches `BellHidden` and `BellFilled`

frontend/packages/client/src/components/Community/SubscribeCommunityButton.js
```diff
@@ -0,0 +1,75 @@
...
+  const containerStyles = sizes[size] ?? { height: 48, maxWidth: 125 };
+
+  return (
+    <div
+      className={`column p-0 is-narrow-tablet is-full-mobile ${className}`}
+      style={containerStyles}
+    >
+      <button className={buttonClasses} onClick={onOpenModal}>
+        <Svg name={buttonIcon} />
+      </button>
```
germanurrus, line 72: on mobile the icon is breaking: 
Screen Shot 2022-10-18 at 12 21 49

mannynotfound: this should fix it https://github.com/DapperCollectives/CAST/pull/650/commits/ef0ce77cf1cf1b7775df604ac30ffb1f73055665

frontend/packages/client/src/components/Community/SubscribeCommunityButton.js
```diff
@@ -0,0 +1,75 @@
...
+
+export default function SubscribeCommunityButton({
+  communityId,
+  className,
+  size = 'small',
+}) {
+  const { openModal, closeModal } = useModalContext();
+  const { notificationSettings } = useNotificationServiceContext();
+  const isSubscribed = notificationSettings?.communitySubscription.some(
+    (c) => c === communityId
```
amyliu6, line 16: here we need check for the higher level user preference `isSubscribedFromCommunityUpdates` too.
```suggestion
    const isSubscribed = notificationSettings.isSubscribedFromCommunityUpdates && <community level Subscription is true>
```
If `isSubscribedFromCommunityUpdates` is false, during subscription it need invoke resubscribe method too to enable Email: https://github.com/DapperCollectives/CAST/blob/8cbc14696f9369c690b31d918f917ad85f035d94/frontend/packages/client/src/contexts/NotificationService.js#L115

mannynotfound: if im understanding correctly this should be the logic we want:

https://github.com/DapperCollectives/CAST/pull/650/commits/e0e75e01547c99707f216cda0239a10d1d7c1ec4

PR #652, opened 2022-10-18
Notification LeanPlum Integration

tuffon:


tuffon:
I removed all context work and simplified to an API wrapper

frontend/packages/client/src/api/constants.js
```diff
@@ -5,3 +5,13 @@ export const COMMUNITIES_URL = `${API_BASE_URL}/communities`;
 export const PROPOSALS_URL = `${API_BASE_URL}/proposals`;
 
 export const IPFS_GETWAY = `${process.env.REACT_APP_IPFS_GATEWAY}`;
+
+export const LP_APP_ID = process.env.REACT_APP_LP_APP_ID;
```
amyliu6, line 9: maybe keep it as `LEANPLUM` instead of `LP` for readability

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
...
+    let userAttributes = {};
+    let communityKey = 'community' + communityId;
+    userAttributes[communityKey] = false;
+    Leanplum.setUserAttributes(userAttributes);
+  } catch (e) {
+    throw new Error(e);
+  }
+};
+
+export const subscribeCommunity = async (communityId) => {
```
amyliu6, line 82: on subscribe user is able to subscribe to All communities they are currently watching, ref: FE PR: https://github.com/DapperCollectives/CAST/pull/632/files#diff-c041fdf295a07a252c342ce2bf4c38cdac20e7e542b2cbe9ef1ac820ee1d9e68R22 and [figma design](https://www.figma.com/file/1GRfRrjNki3vR4T1LfLkHi/CAST%3A-Explorations?node-id=501%3A6698)

Might need to extend this method to support subscribe to all communities by fetching list of community ids that user is watching.

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
...
+            subscribed: data.userAttributes[property] === 'True',
+          });
+        }
+      }
+
+      res.communitySubscription = communitySubscriptions;
+
+      if (data.unsubscribeCategories) {
+        data.unsubscribeCategories.forEach((category) => {
+          if (category.id == 1) {
```
amyliu6, line 49: suggest to use the defined constant here. 
`category.id === COMMUNITY_UPDATES_CATEGORY_ID`

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
+import {
+  LP_APP_ID,
+  LP_CRO_KEY,
+  LP_EXPORT_KEY,
+} from 'api/constants';
+import Leanplum from 'leanplum-sdk';
```
amyliu6, line 6: where is this setup code: https://docs.leanplum.com/reference/javascript-setup#step-2-add-leanplum-to-your-web-application ?

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
...
+            communityId,
+            subscribed: data.userAttributes[property] === 'True',
+          });
+        }
+      }
+
+      res.communitySubscription = communitySubscriptions;
+
+      if (data.unsubscribeCategories) {
+        data.unsubscribeCategories.forEach((category) => {
```
linxiao-li, line 48: Suggest using Array.some()

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
...
+      options
+    ).then((response) => {
+      let data = response.response[0];
+      let communitySubscriptions = [];
+      let isSubscribedToCommunityUpdates = true;
+      const res = {
+        email: data.userAttributes.email,
+      };
+
+      for (const property in data.userAttributes) {
```
linxiao-li, line 35: Nit, prefer use javascript high order functions instead of iterators
`const communitySubscription = data.userAttributes.entries().filter().map()`

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
...
+    Leanplum.setUserAttributes({ email });
+  } catch (e) {
+    throw new Error(e);
+  }
+};
+
+export const unsubscribeCommunity = async (communityId) => {
+  try {
+    let userAttributes = {};
+    let communityKey = 'community' + communityId;
```
linxiao-li, line 74: Doesn't seem like the references are being reassigned, any reason for using let instead of const?

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
...
+    userAttributes[communityKey] = false;
+    Leanplum.setUserAttributes(userAttributes);
```
linxiao-li, line 76:
```suggestion
    Leanplum.setUserAttributes({[`community${communityId}`]:false});
```

linxiao-li (commented):
There are a couple issues around the methods that use fetch:
1. we probably don't need the try catch block if the intention here is to catch the api call errors,  try catch works with async calls using the async/awaits syntax sugar.

3. the return statement in the then block only returns result to its upper scope, aka the fetch api call, we need to return the fetch Promise object in order for the caller to consume the result.

5. We need error handler for fetch before consuming the response since fetch only rejects network errors, any 400, 404 won't go into the catch block. Here's a detailed blog post: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+};
+
+export const getUserSettings = async (walletId) => {
+  try {
+    fetch(
+      `https://api.leanplum.com/api?appId=${LP_APP_ID}&clientKey=${LP_EXPORT_KEY}&apiVersion=1.0.6&userId=${walletId}&action=exportUser`,
+      options
+    ).then((response) => {
+      let data = response.response[0];
```
linxiao-li, line 28: This looks a little odd to me, the Response constructor doesn't contain a response property according to the documentation, https://developer.mozilla.org/en-US/docs/Web/API/Response

linxiao-li, line 28:
```suggestion
      const data = response.response[0];
```

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
...
+    fetch(
+      `https://www.leanplum.com/api?appId=${LP_APP_ID}&clientKey=${LP_CRO_KEY}&action=setUserAttributes&userId=${walletId}&unsubscribeCategoriesToAdd=${COMMUNITY_UPDATES_CATEGORY_ID}`,
+      options
+    ).then((response) => {
+      return response;
+    });
+  } catch (e) {
+    throw new Error(e);
+  }
+};
```
germanurrus, line 104: I think the fetch request might need a tweak: 
The Promise returned from fetch() won't reject on HTTP error status even if the response is an HTTP 404 or 500. Instead, as soon as the server responds with headers, the Promise will resolve normally (with the ok property of the response set to false if the response isn't in the range 200–299), and it will only reject on network failure or if anything prevented the request from completing.


```
try {
  const response = await fetch('....');
  if (!response.ok) {
    // request didn't work
    // throw an error 
  }
return response;
} catch (error) {
  console.log('there is an error  ');
  // do something with the error but do not bubble it up otherwise do not catch it and leave the caller function to handle it
}
```

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,117 @@
...
+  try {
+    Leanplum.setUserId(walletId);
+  } catch (e) {
+    throw new Error(e);
+  }
+};
+
+export const getUserSettings = async (walletId) => {
+  try {
+    fetch(
```
germanurrus, line 24: same here, check on response.ok before doing something

frontend/packages/client/src/api/notificationService.js
germanurrus: something like this?
```
import { IS_LOCAL_DEV } from 'const';

if (IS_LOCAL_DEV) {
 Leanplum.setAppIdForDevelopmentMode("YOUR_APP_ID", "YOUR_DEVELOPMENT_KEY");
} else {
 Leanplum.setAppIdForProductionMode("YOUR_APP_ID", "YOUR_PRODUCTION_KEY");
}

there's also a .start() function that needs to be called as well

**Review by germanurrus — CHANGES_REQUESTED** [review]
added some comments

File: frontend/packages/client/src/api/notificationService.js (hunk 2 above)
Comment (tuffon) [reply frontend/packages/client/src/api/notificationService.js:82]: This was not included in the design and i'm not convinced it is necessary, do you want to discuss to clarify what the need is here for me?

Comment (amyliu6) [reply frontend/packages/client/src/api/notificationService.js:82]: as discussed, "subscribe to All" has been descoped, it's ok to leave it as it is for now.

File: frontend/packages/client/src/api/notificationService.js (hunk 9 above)
Comment (tuffon) [reply frontend/packages/client/src/api/notificationService.js:28]: after you transform the response body stream to json then it takes on this object result array shape

File: frontend/packages/client/src/api/notificationService.js [hunk 12]
```diff
@@ -0,0 +1,132 @@
...
+        if (data.unsubscribeCategories) {
+          data.unsubscribeCategories.forEach((category) => {
+            if (parseInt(category.id) === COMMUNITY_UPDATES_CATEGORY_ID) {
+              isSubscribedToCommunityUpdates = false;
+            }
+          });
+        }
+
+        res.isSubscribedToCommunityUpdates = isSubscribedToCommunityUpdates;
+        return res;
```
linxiao-li, line 65: The caller won't be able to access this returned value, checkout this example I put together: https://codepen.io/linxiaodapper/pen/poKozGv?editors=0012

linxiao-li (changes requested):
Let me know if you want to pair if any comment doesn't make sense

frontend/packages/client/src/contexts/NotificationService.js
```diff
@@ -28,6 +34,16 @@ const updateCommunitySubscriptionState = (
...
 
+const IS_LOCAL_DEV = process.env.REACT_APP_APP_ENV === 'development';
+
+if (IS_LOCAL_DEV) {
+  Leanplum.setAppIdForDevelopmentMode(LEANPLUM_APP_ID, LEANPLUM_DEV_KEY);
+} else {
+  Leanplum.setAppIdForProductionMode(LEANPLUM_APP_ID, LEANPLUM_PROD_KEY);
+}
+
+Leanplum.start();
```
mannynotfound, line 45: not saying this is the case, but i've had issues with other API libraries where calling `.start()` from the top level can cause the app to crash/error if that underlying api service is down,. Generally I think it's better to start the service only when you need it (when a user triggers a part of the app that explicitly needs leanplum to render/work properly)

frontend/packages/client/src/App.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
mannynotfound, line 1: what is this needed for? looks like maybe used for testing but should be removed

frontend/package.json
```diff
@@ -12,7 +12,9 @@
       "packages/*"
     ]
   },
-  "dependencies": {},
+  "dependencies": {
+    "leanplum-sdk": "^1.10.5"
+  },
```
mannynotfound, line 17: this should be in `packages/client/package.json`

frontend/packages/client/src/api/notificationService.js
tuffon: good catch here

frontend/packages/client/src/contexts/NotificationService.js
tuffon: We need to run this once when the app starts, we can potentially move it to the login event, can you suggest a file/location to move this code to?

frontend/packages/client/src/api/notificationService.js
```diff
@@ -0,0 +1,129 @@
...
+    Leanplum.setUserAttributes({ [`community${communityId}`]: false });
+    return true
+  } catch (e) {
+    throw new Error(e);
+  }
+};
+
+export const subscribeCommunity = async (communityId) => {
+  try {
+    Leanplum.setUserAttributes({ [`community${communityId}`]: true });
```
amyliu6, line 90: as per the [doc](https://docs.leanplum.com/reference/user-attributes) only `string` and `number` type is supported, it looks like when fetch the data from Leanplum it will return as "True" ? https://github.com/DapperCollectives/CAST/pull/652/files#diff-57d84cd560a449e825c5114c1a3bf8924c84d9ea141de174a3b3555be6b020a3R48

Maybe we can explicitly store it as "true" to avoid implicit data conversion.

tuffon: it accepts this value but will return as a string, I have tested both cases and they both work

amyliu6: i see, I would suggest setting the value explicitly to be a string as per their doc, that way we can expect the same value in get with no questions. (today Leanplum implicitly convert boolean to "True", we won't know how they will treat unsupported type in future).

frontend/packages/client/src/contexts/NotificationService.js
mannynotfound: @tuffon having it called on login makes sense to me since we wont be able to get a user's preferences without that anyway 👍

tuffon: I'm not familiar with where that logic should be placed can you suggest a location? otherwise I will keep it here

linxiao-li: Can we wrap this section of code up into a `startLeanplum` and put it into `api/notificationService.js`? So the implementation details of leanPlum can be encapsulated.

tuffon: sure, when you integrate you can put the function call where you need it

amyliu6 (approved):

linxiao-li (approved):

PR #655, opened 2022-10-18
VSL-245: add toast component

mannynotfound:
- adds `@chakra-ui/react` and its dependencies
- create `useToastHook` that wraps around chakra's `useToast` to render a custom "CAST" style toast design

## Usage:

```js
const { popToast } = useToast();
// ....
popToast("My toast message", {
  footerText: "My toast link", // optional
  footerLink: "/my-path", // optional ,only used with footerText
});
```

image


frontend/packages/client/package.json
```diff
@@ -3,8 +3,11 @@
   "version": "1.0.1",
   "private": true,
   "dependencies": {
+    "@chakra-ui/react": "1.8.9",
```
mannynotfound, line 6: use v1 of chakra as v2 only supports React 18

frontend/packages/client/package.json
```diff
@@ -43,7 +47,7 @@
     "react-hook-form": "^7.11.0",
     "react-images-upload": "^1.2.8",
     "react-router-dom": "^5.2.0",
-    "react-scripts": "4.0.3",
+    "react-scripts": "5.0.1",
```
mannynotfound, line 50: upgrade to react scripts 5 because of this bug: https://github.com/framer/motion/issues/1525

also semi-related I sometimes get this bug with v4 that is mad annoying https://stackoverflow.com/questions/69051008/react-injecting-iframe-with-max-z-index-on-reload-after-changes-development

frontend/packages/client/src/App.js
```diff
@@ -31,13 +32,15 @@ function App() {
...
-            </NotificationServiceProvider>
+            <ChakraProvider>
+              <NotificationServiceProvider>
+                <NotificationModalProvider>
+                  <ErrorHandler>
+                    <AppPages />
+                  </ErrorHandler>
+                </NotificationModalProvider>
+              </NotificationServiceProvider>
+            </ChakraProvider>
```
mannynotfound, line 43: our app.js is starting to look like this but its OK 

image

frontend/packages/client/src/components/common/Toast.js
```diff
@@ -0,0 +1,43 @@
...
+    >
+      <div
+        className="cursor-pointer"
+        onClick={onClose}
+        style={{
+          position: 'absolute',
+          top: 8,
+          right: 8,
+        }}
+      >
```
mannynotfound, line 20: trying to use less bulma so went with absolute position close vs the `columns` spaghetti we usually have to create

frontend/packages/client/src/components/common/Toast.js
```diff
@@ -0,0 +1,43 @@
...
+        <Svg name="Close" width="18" heigth="18" />
+      </div>
+      <div>{text}</div>
+      {footerText && (
+        <div
+          className="mt-1"
+          onClick={() => {
+            if (footerLink) {
+              history.push(footerLink);
```
mannynotfound, line 29: have to use `history` here instead of `<Link>` because the ChakraToast wrapper injects itself outside of the app and thus doesnt have the Router context

frontend/packages/client/.env.development
```diff
@@ -0,0 +1 @@
+GENERATE_SOURCEMAP=false
```
mannynotfound, line 1: this is needed to avoid ongoing bug: https://github.com/facebook/create-react-app/pull/11752

germanurrus:
@mannynotfound Would it be possible to merged this with release-v1.6.0? so we can start incorporating chakra?

mannynotfound:
@germanurrus sure! opened here https://github.com/DapperCollectives/CAST/pull/659

PR #656, opened 2022-10-18
CAS-555 & CAS-556: Create endpoint to allow saving drafts

0xmovses:
• updates proposal controllers helpers and adds new `proposal.go` model methods for CRUDing Draft Proposals.  
• adds tests for various cases. 
• adds `errProposalNotFound` type for clearer error response.

### DIFF
--- a/backend/main/server/helpers.go
+++ b/backend/main/server/helpers.go
@@ -706,7 +706,7 @@ func handlePermissionErrorr(result models.CanUserCreateProposalResponse) errorRe
 }
 
 func (h *Helpers) validateUpdateProposalStatus(status string) error {
-	validStatuses := []string{"draft", "pending", "published", "cancelled"}
+	validStatuses := []string{"draft", "closed", "published", "cancelled"}
 
 	for _, validStatus := range validStatuses {
 		if status == validStatus {
--- a/backend/main/server/routes.go
+++ b/backend/main/server/routes.go
@@ -35,7 +35,6 @@ func (a *App) initializeRoutes() {
 	a.Router.HandleFunc("/proposals/{proposalId:[0-9]+}/votes", a.createVoteForProposal).Methods("POST", "OPTIONS")
 	a.Router.HandleFunc("/votes/{addr:0x[a-zA-Z0-9]+}", a.getVotesForAddress).Methods("GET")
 	//Strategies
-	// a.Router.HandleFunc("/proposals/{proposalId:[0-9]+}/votes/{addr:0x[a-zA-Z0-9]{16}}", a.updateVoteForProposal).Methods("PUT", "OPTIONS")
 	a.Router.HandleFunc("/proposals/{proposalId:[0-9]+}/results", a.getResultsForProposal)
 	// Types
 	a.Router.HandleFunc("/voting-strategies", a.getVotingStrategies).Methods("GET")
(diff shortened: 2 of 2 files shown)


jbluks (approved):
One test appears to be failing:

Screen Shot 2022-10-20 at 9 16 28 AM

backend/main/models/proposal.go
```diff
@@ -176,6 +188,31 @@ func (p *Proposal) UpdateProposal(db *s.Database) error {
...
+		choices = COALESCE($2, choices),
+		strategy = COALESCE($3, strategy),
+		min_balance = COALESCE($4, min_balance),
+		max_weight = COALESCE($5, max_weight),
+		start_time = COALESCE($6, start_time),
+		end_time = COALESCE($7, end_time),
+		body = COALESCE($8, body),
+		block_height = COALESCE($9, block_height),
+		cid = COALESCE($10, cid),
+		composite_signatures = COALESCE($11, composite_signatures),
```
jacksonConrad, line 204: i wonder if we should be updating composite_signatures/voucher fields.... in the long run it would be nice to collect and store a signature for each change that is made to the proposal to have a proper audit trail, but in lieu of that, we should just save the initial signature/voucher that was submitted to create the draft proposal in the first place, so its consistent.

obviously we should still validate the signature/voucher in the update API, but i dont think we should overwrite the original signatures in the DB

jacksonConrad (commented):
Looking at this PR, its not clear to me how a user would go about moving a proposal from "draft" to "pending"/"published".  Is there a way to do this?

Also, my first instinct would probably have been to roll the "draft" proposal functionality into the existing Proposals REST API, rather than creating a new set of endpoints.  The way it is right now, it looks like you could call these `/proposals/draft` endpoints on non-draft proposals and they would still work (i don't see any validations asserting that a proposal actually has the "draft" status before deleting it, for instance).  Rolling it into the original proposal REST API might make it easier to do explicit separation of concerns based on the proposal's status.

backend/tests/test_utils/proposal_utils.go
```diff
@@ -138,16 +116,37 @@ func (otu *OverflowTestUtils) GenerateDraftProposalStruct(
...
 	proposal.Start_time = time.Now().AddDate(0, 1, 0)
 	proposal.End_time = time.Now().Add(30 * 24 * time.Hour)
-
-	proposal.Status = nil
 	return &proposal
 }
 
 func (otu *OverflowTestUtils) GenerateCancelProposalStruct(
 	signer string,
 	proposalId int,
```
jacksonConrad, line 124: looks like proposalId isn't being used in this function or the two below. we should either include it in the `Proposal` model, or omit it from the args

backend/main/models/proposal.go
```diff
@@ -200,11 +199,20 @@ func (p *Proposal) UpdateDraftProposal(db *s.Database) error {
 		end_time = COALESCE($7, end_time),
 		body = COALESCE($8, body),
 		block_height = COALESCE($9, block_height),
-		cid = COALESCE($10, cid),
-		composite_signatures = COALESCE($11, composite_signatures),
-		voucher = COALESCE($12, voucher)
-		WHERE id = $13
-	`, p.Name, p.Choices, p.Strategy, p.Min_balance, p.Max_weight, p.Start_time, p.End_time, p.Body, p.Block_height, p.Cid, p.Composite_signatures, p.Voucher, p.ID)
```
jacksonConrad: maybe this is another ticket, but doesnt look like there is currently a way to take a proposal from "draft" to "published"... should `status` be included as part of `UpdateDraftProposal`, or should it be a separate endpoint/action?

backend/main/server/controllers.go
```diff
@@ -512,16 +432,52 @@ func (a *App) updateProposal(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	p.Status = &payload.Status
-	p.Cid, err = helpers.pinJSONToIpfs(p)
+	if *payload.Proposal.Status == "draft" {
```
jacksonConrad, line 435: should check the actual status of the proposal in the DB, not the status submitted in the payload since it could be anything

backend/main/server/controllers.go
```diff
@@ -512,16 +432,52 @@ func (a *App) updateProposal(w http.ResponseWriter, r *http.Request) {
...
+
+		if err := p.UpdateDraftProposal(a.DB); err != nil {
+			log.Error().Err(err).Msg("Error updating draft proposal")
+			respondWithError(w, errIncompleteRequest)
+			return
+		}
+
+		respondWithJSON(w, http.StatusOK, p)
+	} else {
+		p.Status = payload.Proposal.Status
```
jacksonConrad, line 446: should add a check here that the status transition is valid. i.e.

1. "draft" -> "published"
2. "published" -> "cancelled"

backend/main/models/proposal.go
jacksonConrad: oh nevermind, i addressed this elsewhere in another comment!

jacksonConrad (changes requested):
Looking good! Just need a little extra validation around proposal status I think

backend/main/server/helpers.go
```diff
@@ -592,6 +596,127 @@ func (h *Helpers) createProposal(p models.Proposal) (models.Proposal, errorRespo
...
+		errPermissionsResponse := errCreateProposalPermissions
+		errPermissionsResponse.Message = result.Reason
+		errPermissionsResponse.Details = result.Error.Error()
+		return errPermissionsResponse
+	}
+	return nilErr
+}
+
+func (h *Helpers) validateUpdateProposalStatus(status string) error {
+	validStatuses := []string{"draft", "pending", "published", "cancelled"}
```
jacksonConrad, line 709: we can remove `pending` here, i dont think its ever used as a `status`, just a `computedStatus`

jacksonConrad (approved):

PR #657, opened 2022-10-18
Typescript

dbslone:


frontend/packages/client/src/components/HomeHeader.tsx
```diff
@@ -10,7 +10,13 @@ const stars = [
...
 ];
 
-export default function HomeHeader({ isVisible = true } = {}) {
+interface HomeHeaderProps {
+  isVisible: boolean;
+}
+
+export default function HomeHeader({
+  isVisible = true,
+}: HomeHeaderProps): JSX.Element | null {
```
germanurrus, line 19: Could we start declaring things like this?: 
```
export const HomeHeader: React.FC<HomeHeaderProps> = ({ isVisible = true }) => {
  return ....;
};
```

FC has null as a return type as well and makes is a bit more readable (IMO), although using JSX.Element we need to be more explicit in the types declaration (stronger types) and that's good too.

germanurrus (approved):
LGTM! 🚀 (left some comments)

frontend/packages/client/src/components/HomeHeader.tsx
dbslone: Yea I'll switch to use React.FC

frontend/packages/client/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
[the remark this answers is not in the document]
dbslone: hmm thats odd the .d.ts file is in the shared-components lib. Sometimes the typescript compiler in vs code does wonky things. If you hit cmd+shift+p then enter typescript you should see a restart option

linxiao-li (commented):
I think it might be helpful to compile a react typescript style guide which include examples of common patterns:
- Functional Components
- Hooks
- Context
- Event Handler

jbluks (approved):

germanurrus (approved):

germanurrus:
@linxiao-li @mannynotfound do you have any comments about getting this PR merged?

linxiao-li (approved):
LGTM

PR #666, opened 2022-10-20
CAS-599: Check if user meets requirements to create a proposal

germanurrus:
- Adds custom theme updates for Chakra
- Implements custom component using Chakra

Screen Shot 2022-10-24 at 19 39 50


frontend/packages/client/src/helpers/theme/index.js
```diff
@@ -1,14 +1,44 @@
...
+    Card: CardStyle,
+    Link,
+    Text,
+    // Other components go here
   },
-});
+};
 
-export default theme;
+export default extendTheme(overrides);
```
germanurrus, line 44: this file provides configuration to use a custom theme with chakra

`definitions.js` file overwrites default definitions by Chakra to match designs
`components` folder contains one file per component we are overwriting to match designs, by having one file per component it's easy to find and keeps things more isolated 
`foundations` folder contains definitions for borders, radius, typography (currently we are using bulma and it should be ported here)
`styles` file defines global styles for html elements
there's more to add but this is the basic structure

frontend/packages/client/src/components/Card/styles.js
```diff
@@ -0,0 +1,37 @@
...
+      minWidth: '210px',
+      height: '69px',
+    },
+  },
+  defaultProps: {
+    background: 'white',
+  },
+};
+
+export default Card;
```
germanurrus, line 37: this is a custom component, therefore defining the style here makes it easier to identify it as a custom component and not mixing this definition with the theme overwrites.

frontend/packages/client/src/components/Card/Card.js
```diff
@@ -0,0 +1,10 @@
+import { Box, useStyleConfig } from '@chakra-ui/react';
+
+export default function Card(props) {
+  const { variant, ...rest } = props;
+
+  const styles = useStyleConfig('Card', { variant });
+
+  // Pass the computed styles into the `__css` prop
+  return <Box __css={styles} {...rest} />;
+}
```
germanurrus, line 10: Custom component

jbluks (approved):
LGTM!

jbluks (changes requested):
Can we confirm that these changes are meant for release-v1.6.0 before merging? (Confirmed)

jbluks (approved):

jacksonConrad (approved):

PR #669, opened 2022-10-24
CAS-654 : additional tests for draft proposals

0xmovses:
• adds two new test cases for draft proposals :

`should be able to update the status of a draft proposal to 'published'`
`should be able to update a draft proposal with a new body`

• adds new `proposals_utils.go` generate functions. 

Side thought for new ticket. Wondering if we can make any of these generate functions more generalised? So to reduce the number of functions here. It's getting a tad unwieldily.


jbluks (approved):
LGTM

PR #674, opened 2022-10-25
CAS-658 & CAS-665: Get Create and Update User Endpoint

0xmovses:
•creates get, update and create User routes.
•creates controller and model methods for the routes above. 
•adds test to cover the basic cases. 

**N.B** 
For image upload, the front end has to first hit `/upload` endpoint. Then with the returned ipfs url, use that as the value for the `profileImage` payload field for `/user` POST endpoint. This was done because to upload an image to `/user` would mean using both form data and json as payload type in one call, which we haven't done anywhere else. 

I haven't added `validateSignature` functions on these routes, as the approve wallet window wasn't indicated in the design prototypes. Do we want this?

Also I added a `uuid` field for the users as primary key. I thought it might be useful at some stage in the future to have our own unique identifier for users aside from addresses (which are out of our control); just a nice to have, but can drop this if we feel it will add clutter.


jbluks (approved):
LGTM

jbluks:
One question I do have - is this part of release-v1.6.0? Or is this part of a separate release?

jbluks (changes requested):
Let's confirm the release branch target before merging.

jbluks (approved):
New Target is v2.0

backend/main/models/user.go
```diff
@@ -0,0 +1,112 @@
+package models
+
+import (
+	"github.com/DapperCollectives/CAST/backend/main/shared"
+	"github.com/google/uuid"
+)
+
+type User struct {
+	Uuid          *string `json:"uuid,omitempty"`
+	Addr          *string `json:"address"`
```
jacksonConrad, line 10: lets make `Addr` field required, so `validatePayload` will at least do something :)

backend/main/server/controllers.go
```diff
@@ -894,6 +901,57 @@ func (a *App) createCommunityUser(w http.ResponseWriter, r *http.Request) {
...
+
+func (a *App) createUser(w http.ResponseWriter, r *http.Request) {
+	payload := models.User{}
+
+	if err := validatePayload(r.Body, &payload); err != nil {
+		log.Error().Err(err).Msg("Error validating payload")
+		respondWithError(w, errIncompleteRequest)
+		return
+	}
+
```
jacksonConrad, line 926: we should definitely require/validate a signature to create & update a user, otherwise the API is wide open.

you can ignore the `Voucher` case, because Dapper Wallet has finally implemented user signing, so that will most likely be going out with this release as well.

jacksonConrad (changes requested):
Looking good!

I do think it makes sense to validate a user signature for the Create & Update flows, though, and give Design/Product a heads-up.

backend/tests/test_utils/user_utils.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+func (otu *OverflowTestUtils) GenerateUserPayload(signer string, user models.User) *models.User {
+	payload := user
+	fmt.Printf("payload: %+v \n", payload)
+	timestamp := fmt.Sprint(time.Now().UnixNano() / int64(time.Millisecond))
+	fmt.Printf("timestamp: %s \n", timestamp)
+	compositeSignatures := otu.GenerateCompositeSignatures(signer, timestamp)
+	fmt.Printf("compositeSignatures: %+v \n", compositeSignatures)
```
dbslone, line 64: remove logs

jacksonConrad (approved):

PR #680, opened 2022-10-26
CAS-673: Fix update proposal payload

jbluks:
The update proposal payload was incorrect and did not provide the payload field containing the updated status.

I am not sure if the fix I made is the correct fix, but it works now. Assuming it worked before, so not sure whether the back-end was updated and that broke it, or the front-end.

### DIFF
--- a/frontend/packages/client/src/api/proposals.js
+++ b/frontend/packages/client/src/api/proposals.js
@@ -87,7 +87,8 @@ export const updateProposalApiReq = async ({
       'Content-Type': 'application/json',
     },
     body: JSON.stringify({
-      ...updatePayload,
+      payload: updatePayload,
+      signingAddr: updatePayload.signingAddr,
       timestamp,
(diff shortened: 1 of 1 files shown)


germanurrus:
@jbluks backend endpoint payload was updated causing the frontend to break: 
This is the PR: https://github.com/DapperCollectives/CAST/pull/656
This is the line that changed: 
https://github.com/DapperCollectives/CAST/pull/656/files#diff-7778dd3d4999c108290a8d41e171d479d56b27c482e31354ef84476a6e2a2095R48

germanurrus (approved):

jbluks:
> how could we have caught this sooner? Backend test code can update without the client knowing. A front end cypress test file just for testing the api payloads?

Yea I think cypress tests of all the core functionality would work because we need to hit the actual backend to find an issue like this. If we were dealing with mocks on the front-end, this would still be missed.

PR #672, opened 2022-10-24
VSL-262 - rework notification manage modal to be a toast instead

mannynotfound:
- delete `modals/Notifications/Manage.js`
- add logic for popping toast in `SubscribeCommunityButton`

image

PR #683, opened 2022-10-27
manny/vsl-266 - add subscribed state to post vote options

mannynotfound:
builds off of https://github.com/DapperCollectives/CAST/pull/672 to add the new `SubscribeCommunityButton` to the `PostVoteOptions` component while also adding a new size / icon to `SubscribeCommunityButton` to match the design

image
image
image


frontend/packages/client/src/components/common/Toast.js
```diff
@@ -1,43 +1,38 @@
-import { useHistory } from 'react-router-dom';
 import { Svg } from '@cast/shared-components';
 
-const Toast = ({ onClose, text, footerText, footerLink }) => {
-  const history = useHistory();
-
-  return (
+const Toast = ({ history, onClose, text, footerText, footerLink }) => (
```
linxiao-li, line 3: Can we adjust the structure of this Toast component to be something like this:
```
<wrapperDiv>
    <closeButton/>
    {children}
</wrapperDiv>

```
Reason: this way we can pass any children components into the toast, have the business logic within their own context, and make the toast component more flexible and reusable

frontend/packages/client/src/hooks/useToast.js
```diff
@@ -1,15 +1,22 @@
+import { useHistory } from 'react-router-dom';
 import Toast from 'components/common/Toast';
 import { useToast } from '@chakra-ui/react';
 
 export default function useToastHook() {
   const toast = useToast();
+  const history = useHistory();
```
linxiao-li, line 7: Here as well, if we have the toast just be a wrapper of whatever content, we have this hook assume the sole responsibility of managing the show and hide of toast component

mannynotfound:
@linxiao-li so after syncing we decided to make Toast take a "type" that dictates the styling / icons and then provide `actionText`/`actionFn` if user needs to take an action. Updated to that while also updating to latest toast Figma design ( https://www.figma.com/file/bni5O6c7Yh1rCUwFRO8f9i/CAST%3A-Design-Library?node-id=348%3A3675 ). Should be a lot cleaner now and look like this:

image

frontend/packages/client/src/components/common/Toast.js
```diff
@@ -1,41 +1,67 @@
...
+    icon: <img src="/spinner.gif" alt="loading spinner" />,
+  },
+  success: {
+    bgColor: '#F5FBF6',
+    borderColor: '#2BA148',
+    icon: <Svg name="CheckMark" />,
+  },
+  info: {
+    bgColor: '#FEF4CD',
+    borderColor: '#C8A104',
```
mannynotfound, line 16: hardcoding these color styles here because theyre only used in the toast (vary slightly from other colors we use) and to avoid more `App.sass` bloat

frontend/packages/client/src/components/common/Toast.js
```diff
@@ -1,41 +1,67 @@
...
+    icon: null,
+  },
+};
+
+const Toast = ({ onClose, message, messageType, actionText, actionFn }) => {
+  const { bgColor, borderColor, icon } =
+    TOAST_PROPS[messageType] ?? TOAST_PROPS.info;
```
linxiao-li, line 23: nit
```suggestion
const Toast = ({ onClose, message, messageType='info', actionText, actionFn }) => {
  const { bgColor, borderColor, icon } =
    TOAST_PROPS[messageType];
```

linxiao-li (approved):
LGTM

PR #684, opened 2022-10-27
manny/vsl-264: standardize input and add red color to border

mannynotfound:
- standardized input classnames that were used in every instance of `Input`: `rounded-sm border-light p-3 column is-full`
-  default to `border-light` unless there is an error in which case use `border-danger`

image

### DIFF
--- a/frontend/packages/client/src/components/common/Input.js
+++ b/frontend/packages/client/src/components/common/Input.js
@@ -1,4 +1,5 @@
 import FadeIn from 'components/FadeIn';
+import classnames from 'classnames';
 
 export default function Input({
   style = {},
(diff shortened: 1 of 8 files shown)


frontend/packages/client/src/components/common/Input.js
```diff
@@ -9,21 +10,31 @@ export default function Input({
   disabled,
   error,
   type = 'text',
-  conatinerClassNames = '',
+  containerClassNames = '',
```
mannynotfound, line 13: fix typo

linxiao-li (commented):
LGTM

germanurrus (approved):
LGTM!

PR #691, opened 2022-10-28
CAS-674: Enable next button on form with dirty field status

germanurrus:
Some field arrays are not validated onChange mode, this enables next button to trigger validation. This happens when switching from/to ranked-vote-choice single choice


jbluks (approved):
LGTM

PR #693, opened 2022-10-28
Job Worker Consumer Producer & Notification Service

tuffon:
* Fixed migrations and bugs with server

  * Added Leanplum API services

  * Some Clean-Up


tuffon:
There is one bug still remaining, I could use a pair on implementing go routines and channels in order to ensure that the main server listen thread is not blocked when the system is sent a SIGINT

backend/main/server/app.go
```diff
@@ -193,6 +193,7 @@ func (a *App) Initialize() {
 func (a *App) Run() {
 	addr := fmt.Sprintf(":%s", os.Getenv("API_PORT"))
 	log.Info().Msgf("Starting server on %s ...", addr)
+	go shared.StartJobWorkers()
```
tuffon, line 196: there is a quirk here where a SIGINT is consumed by the worker and the main server remains running, looking for a solution on how to couple the listeners to close and open together.  Researching more into go routine channels for this, pairing is welcomed

backend/main/shared/notification-service.go
```diff
@@ -0,0 +1,175 @@
...
+	for {
+		fileResponse = getExportFiles(jobId)
+		if fileResponse["state"] == "RUNNING" {
+			time.Sleep(5 * time.Second)
+		} else {
+			log.Info().Msgf("LP Export Job Complete with Response: %+v", fileResponse)
+			break
+		}
+	}
+	fileArr := fileResponse["files"].([]interface{})
```
tuffon, line 53: not a huge fan of casting this way but the leanplum api can return different response object shapes depending on the result, so it was more challenging to define response type structs

PR #694, opened 2022-10-31
CAS-664: WIP User Profile Page Layout

germanurrus:


frontend/packages/client/src/components/ShareDropdown.tsx
```diff
@@ -0,0 +1,96 @@
...
+            >
+              Tweet
+            </Link>
+          </MenuItem>
+        </MenuList>
+      </Menu>
+    </>
+  );
+};
+export default ShareDropdown;
```
germanurrus, line 96: This component is reused in the proposal copy and in the user profile

frontend/packages/client/src/components/Tooltip.tsx
```diff
@@ -23,4 +31,6 @@ export default function Tooltip({
       {children}
     </span>
   );
-}
+};
+
+export default Tooltip;
```
germanurrus, line 36: Turning into TS

frontend/packages/client/src/helpers/theme/components/Button.ts
```diff
@@ -0,0 +1,22 @@
...
+      height: 10, // 2.5rem => 40px
+    },
+    sm: {
+      height: 8, //2.125 => 34px
+    },
+  },
+  defaultProps: {},
+};
+
+export default Button;
```
germanurrus, line 22: adds sizes for buttons based on designs

frontend/packages/client/src/components/CustomTab.tsx
```diff
@@ -0,0 +1,34 @@
...
+import {
+  Flex,
+  HStack,
+  Tab,
+  useMultiStyleConfig,
+  useTab,
+} from '@chakra-ui/react';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const CustomTab = forwardRef<HTMLElement, any>((props, ref) => {
```
dbslone, line 12: Should we add an interface above to define the props for the component instead of using any?

frontend/packages/client/src/components/CustomTab.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <Tab __css={styles.tab} {...tabProps}>
+      <HStack spacing={2}>
+        <Flex {...(isSelected ? { fontWeight: 'bold' } : {})}>
+          {tabProps.children}
+        </Flex>
+        {isSelected ? (
+          <Svg name="Star" width="13" height="13" fill="black" />
+        ) : null}
```
dbslone, line 28: Instead of ternary here you could use `isSelected && <Svg />`

frontend/packages/client/src/components/Proposal/ShareProposalDropdown.tsx
```diff
@@ -0,0 +1,39 @@
...
+    userVoted,
+  } = {} as ShareProposalDropdownProps
+) => {
+  const proposalUrl = `${FRONTEND_URL}/#/community/${communityId}/proposal/${proposalId}`;
+
+  const twitterPost = `https://twitter.com/intent/tweet?text=${encodeURIComponent(
+    userVoted
+      ? `I just voted on ${proposalName} on CAST! ${proposalUrl}`
+      : `Check out ${proposalName} on CAST! ${proposalUrl}`
+  )} `;
```
dbslone, line 25: This should just be
```
const twitterPost = userVoted
      ? `I just voted on ${proposalName} on CAST! ${proposalUrl}`
      : `Check out ${proposalName} on CAST! ${proposalUrl}`
```

because below in the component you using the twitter intent url again

frontend/packages/client/src/components/ShareDropdown.tsx
```diff
@@ -0,0 +1,96 @@
...
+                fontSize={isBiggerThanMobile ? 'md' : 'sm'}
+              >
+                Share
+              </Text>
+            ) : null}
+          </Flex>
+        </MenuButton>
+        <MenuList borderRadius="2xl" maxW="192px" minW="192px">
+          <MenuItem display="flex" alignItems="center" minHeight="48px" pl={3}>
+            <CopyToClipboard text={copyString} onCopy={() => {}}>
```
dbslone, line 67: Maybe we should utilize `useClipboard` from [Chakra](https://chakra-ui.com/docs/hooks/use-clipboard) instead this will allow us to remove `react-copy-to-clipboard`

frontend/packages/client/src/components/UserProfile/ShareProfileDropdown.tsx
```diff
@@ -0,0 +1,27 @@
...
+interface ShareProfileDropdownProps {
+  userAddr: string;
+}
+const ShareProfileDropdown: React.FC<ShareProfileDropdownProps> = ({
+  userAddr,
+}) => {
+  const profileUrl = `${FRONTEND_URL}/#/profile?addr=${userAddr}`;
+
+  const twitterPost = `https://twitter.com/intent/tweet?text=${encodeURIComponent(
+    `Check at my profile in ${profileUrl} on CAST! `
```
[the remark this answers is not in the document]
dbslone: Same thing here as well I don't think you need the url in the string because `ShareDropdown` uses the twitter url and encodes the string passed to the component

frontend/packages/client/src/components/WalletAddress.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <Tooltip
+      classNames="is-flex is-flex-grow-1 is-align-items-center transition-all"
+      position="top"
+      text="Copied!"
+      alwaysVisible={true}
+      enabled={addressCopied}
+    >
+      <CopyToClipboard text={addr} onCopy={markAddressCopied}>
```
dbslone, line 30: Same thing here maybe we can utilize `useClipboard` from Chakra

frontend/packages/client/src/pages/UserProfile.tsx
```diff
@@ -0,0 +1,132 @@
...
+  TabPanels,
+  Tabs,
+  VStack,
+} from '@chakra-ui/react';
+import PageContainer from 'layout/PageContainer';
+
+const UserProfile: React.FC = () => {
+  const {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
dbslone, line 29: if you use `// @ts-expect-error: Enter reason here` then you don't need the eslint disable and we will have a documented reason why for why we need to ignore the line

PR #708, opened 2022-11-01
CAS-698: Fix Voting Success Modal appearing too early

jbluks:


frontend/packages/client/src/pages/Proposal.js
```diff
@@ -298,6 +299,14 @@ export default function ProposalPage() {
...
   }
 
+  console.log(
+    confirmingVote,
+    castingVote,
+    castVote,
+    String(castVote),
+    confirmingVote && castingVote && String(castVote)
+  );
+
```
germanurrus, line 309: remove output

frontend/packages/client/src/pages/Proposal.js
```diff
@@ -162,6 +162,7 @@ export default function ProposalPage() {
   };
 
   const onConfirmVote = (options) => {
+    console.log('onConfirmVote', options);
```
germanurrus, line 165: remove output

frontend/packages/client/src/pages/Proposal.js
```diff
@@ -325,7 +334,7 @@ export default function ProposalPage() {
           </div>
         </div>
       )}
-      {confirmingVote && castingVote && String(castVote) && (
+      {confirmingVote && castingVote && castVote && (
```
germanurrus, line 337: maybe doing `castVote !== null` since it's initialised with null ? an option could be a 0

germanurrus (approved):

frontend/packages/client/src/pages/Proposal.js
jbluks: Yes - that is the correct fix here.

PR #715, opened 2022-11-01
add chakra plugin

germanurrus:
- Adds chakra plugin into shared-components lib
- Moving theme configuration to shared-components lib 
- Added basic story for custom Card component


dbslone (approved):
LGTM

